Issue #80, opened 2014-07-15
MomentFormatConverter confusion


inssein:
I was working on https://github.com/sonata-project/SonataCoreBundle/pull/64 and I didn't like how the MomentFormatConverter worked because I ended up adding a lot of mappings.

I want to improve the class, and this can be easily done by creating a mapping between format character in PHP to format character(s) in moment.js (i.e: 'H' => 'HH').

The MomentFormatConverter states that it stores mapping between PHP and moment.js, but that makes no sense because https://github.com/sonata-project/SonataCoreBundle/blob/master/Date/MomentFormatConverter.php#L38 is not a valid PHP format string. All PHP format characters are only one character in length.

Once I know what MomentFormatConverter is really trying to do, I can try to improve it.

Issue #85, opened 2014-07-31
DateTime issue with momentjs


rande:
Hello,

I have reverted @na-ji works has some commit breaks:
- [x] https://github.com/sonata-project/SonataCoreBundle/issues/80
- [x] https://github.com/sonata-project/SonataAdminBundle/issues/2261
- [x] https://github.com/sonata-project/SonataAdminBundle/issues/2255
- [x] https://github.com/sonata-project/SonataAdminBundle/issues/1887
- [x] https://github.com/sonata-project/sandbox/issues/400

Related commits:
- 294d70b79f657d125f7c26b8d1f8402185c442c6
- 6af1033009c7fc0dafbc11fdc98aca8b07db775b
- 1f7b88fb80f135d286d281537fb54a48b29b807a

There is also others reference of this issue in the SonataAdminBundle.

AntonioCS:
Has this been adopted by the bundle? I didn't set anything but the element with sonata_type_datetime_picker has the attribute 'data-date-format' => 'YYYY-MM-DDTHH:mm:ssZZ'.
This looks horrible to the user

debach:
In order to change the format and the language of the `sonata_type_datetime_picker`, I had to set the following options,

    ->add('when', 'sonata_type_datetime_picker', array(
              'dp_language' => 'de',
              'format' => 'dd.MM.yyyy, HH:mm',
              'attr' => array(
                  'data-date-format' => 'DD.MM.YYYY, HH:mm',
              ))

and load the moment.js locale manually by

    {% javascripts
         '@SonataCoreBundle/Resources/public/vendor/moment/locale/de.js'
    %}
        <script src="{{ asset_url }}"></script>
    {% endjavascripts %}

Is this intended? It’s not at all obvious from the documentation.

AntonioCS:
@debach commented on [29 Jan 2015 10:44 GMT](https://github.com/sonata-project/SonataCoreBundle/issues/85#issuecomment-72004662 "2015-01-29T10:44:27Z - Replied by Github Comment Enhancer"):
> In order to change the format and the language of the `sonata_type_datetime_picker`, I had to set the following options,
> 
>  -&gt;add('when', 'sonata_type_datetime_picker', array(
>            'dp_language' =&gt; 'de',
>            'format' =&gt; 'dd.MM.yyyy, HH:mm',
>            'attr' =&gt; array(
>                'data-date-format' =&gt; 'DD.MM.YYYY, HH:mm',
[...]

PR #93, opened 2014-08-16
Two new timestamp formats added to MomentFormatConverter

caponica:
Also see my comments on #80


webdevilopers:
Is this PR still up to date @caponica ? Could you please take a look at the `german date` format I suggested in https://github.com/sonata-project/SonataCoreBundle/issues/85#issuecomment-74664411?

Thanks

Bladrak:
Seems good to me, @rande could you check it out?

PR #125, opened 2015-01-19
Update datetimepicker format

StudioMaX:
In latest version DateTimePicker doesn't work when enabled 'datepicker_use_button'.
Also in latest version 'data-date-format' parameter should be in the < input>, not in the < div> (See comit Eonasdan/bootstrap-datetimepicker@38a662c).


StudioMaX:
ping @rande

PR #136, opened 2015-02-03
[Export] Added csv customizable delimiter, fixes #2707

akhairallah:


pulzarraider:
@akhairallah adding new parameter `$delimiter` to `getResponse` method is the easiest solution, but I don't like it.

Instead of adding new parameter that can be used only for CSV export we should move the writers to standard Symfony services. Their constructor parameters can be overloadable parameters.

rande:
This class is just a quick factory to export data. You can reuse the code to build your own.

Issue #138, opened 2015-02-18
Update eonasdan-bootstrap-datetimepicker to current version 4.0


webdevilopers:
As mentioned by @afurculita in issue https://github.com/sonata-project/SonataCoreBundle/issues/115#issuecomment-72275920 version *4.0* of the `eonasdan-bootstrap-datetimepicker` has been released:
- https://github.com/Eonasdan/bootstrap-datetimepicker/issues/161

This version includes new features like the `calendarWeek` options:
- https://github.com/sonata-project/SonataCoreBundle/issues/115

But they also mention some BC:
- https://github.com/Eonasdan/bootstrap-datetimepicker/releases/tag/v4.0.0
- http://eonasdan.github.io/bootstrap-datetimepicker/Version%204%20Changelog/

Can someone tell if these BC are relevant to the `SonataCoreBundle` ?

If not I could create the PR for the `bower.json` file:
https://github.com/sonata-project/SonataCoreBundle/blob/master/bower.json#L8

Changing dependencies from `"eonasdan-bootstrap-datetimepicker": "3.1.3"` to `"eonasdan-bootstrap-datetimepicker": "4.0.0"`.

Not sure if the version by @Bladrak is a better solution linking the `latest` version:
- https://github.com/sonata-project/SonataCoreBundle/blob/2.2/bower.json

Then I would create a PR on the docs for the new options.

webdevilopers:
Just a short note to myself for the available options:
- http://eonasdan.github.io/bootstrap-datetimepicker/Options/
- http://sonata-project.org/bundles/core/master/doc/reference/form_types.html#sonata-type-date-picker-and-sonata-type-datetime-picker
- https://github.com/sonata-project/SonataCoreBundle/blob/master/Form/Type/BasePickerType.php#l80-104

Bladrak:
Not sure about the BC breaks, I guess we'll have to try it out. Feel free to submit a PR with the updated version if you'd like :)

nlzet:
Did you continue with this? I tried setting up v4 of the datetimepicker, but this won't work out of the box.

And there will be BC breaks, for example in the current way there is a option <code>dp_pick_time</code> to enable/disable the time. In v4 this is defined by checking if there are time units in your date format. ()

Other options have other names or work slightly differently and have different default values like null/false.

How would you like to continue with this ? Should a PR with BC breaks be created ?

webdevilopers:
At least we need a full list of options that have changed.

[...]

PR #149, opened 2015-03-12
[WIP] Eonasdan datetimepicker update v4.7.14

nlzet:
See #138 for the previous conversation. Here is a  table the datepicker option changes 

| Option | Changes |
| --- | --- |
| `dp_max_date` | default value changed from `null` to `false` |
| `dp_default_date` | default value changed from `''` to `false` |
| `dp_disabled_dates` | default value changed from `array()` to `false` |
| `dp_enabled_dates` | default value changed from `array()` to `false` |
| `dp_icons` | default value changed with extra icons for `previous`, `next`, `today` and `clear` |
| `dp_show_today` | renamed to `dp_show_today_button` |
| `dp_language` | renamed to `dp_locale` |
| `dp_minute_stepping` | renamed to `dp_stepping` |
| `dp_use_minutes` | removed |
| `dp_use_seconds` | removed |
| `dp_pick_time` | removed, this is no longer a boolean option, this is now defined by the date format. |

For the `dp_pick_time` I now used the $this->getName() check to enable/disable the time picker. I think this shouldn't be defined by an option, but by the form type ?


Form/Type/BasePickerType.php
```diff
@@ -87,20 +84,23 @@ protected function getCommonDefaults()
         return array(
             'widget'                   => 'single_text',
             'datepicker_use_button'    => true,
-            'dp_pick_time'             => true,
             'dp_use_current'           => true,
             'dp_min_date'              => '1/1/1900',
-            'dp_max_date'              => null,
-            'dp_show_today'            => true,
```
rande: we cannot remove the value, we need to be BC compatible. Just convert the values.

nlzet: Do you mean I have to keep using the option `dp_show_today` and convert it to `dp_show_today_button` ?

What about the removed options `dp_use_minutes`, `dp_use_seconds` and `dp_pick_time` ?

nlzet:
ping @rande, could you give some more details about my question, so I can finish this PR.

webdevilopers:
With the current `dev-master` the `dp_use_seconds` already has no effect when setting it on the field:
```php
$formMapper
    ->add('bundleStatus', 'sonata_type_datetime_picker', array(
        'dp_side_by_side' => true,
       'dp_use_seconds' => false
    ));
```

The generated html includes the option but set to `true`:
```javascript
 jQuery(function ($) {
$('#dtp_s5512b911e3621_bundleStatus').datetimepicker({"pickTime":true,"useCurrent":true,"minDate":"1\/1\/1900","maxDate":null,"showToday":true,"language":"de","defaultDate":"","disabledDates":[],"enabledDates":[],"icons":{"time":"glyphicon glyphicon-time","date":"glyphicon glyphicon-calendar","up":"glyphicon glyphicon-chevron-up","down":"glyphicon glyphicon-chevron-down"},"useStrict":false,"daysOfWeekDisabled":[],"useMinutes":true,"minuteStepping":1,"sideBySide":true,"useSeconds":true});
}); 
```

https://github.com/sonata-project/SonataCoreBundle/blob/master/Form/Type/BasePickerType.php#L57

I guess this issue is no longer relevant when this PR is merged @nlzet ?
How to disable `seconds` then?

nlzet:
@webdevilopers Hmm you are right, this option is always automatically set, so specifying it has no use. 
Disabling seconds or minutes in the new version is done by the date / time format

But options like `dp_use_minutes` and `dp_pick_time` are a different story. I can keep them for to prevent BC errors, but they won't have any use.

I will make an update so the renamed options will be transformed to the new options.

webdevilopers:
Thanks @nlzet .

When using the time format instead will a data transformer automatically add seconds as `:00` to a field in order to save it as a full `DATETIME` value in database?

nlzet:
@webdevilopers, the format is now correctly set on the datepicker. I think this library will create BC issues because the datepicker and datetimepicker now actually are the same, the type no longer defines whether to use time or not. Should I continue?

Also there is a version 4.0.0 and 4.7.14 of the plugin, I now updated it to the second one.

Do you have an idea how sonata works with the locale of a date? In my current implementation, when using dp_locale "nl" it fails to save some dates

webdevilopers:
To be honest I have no idea. I guess @rande can tell you.

pborreli:
what is the status of this PR ? I have issue with actuel one too.

nlzet:
Well I didn't continue and I don't know if I should. There are new features but there are a lot of BC issues. Maybe this should be targeted for a new version.

What do you think @rande ?

Soullivaneuh:
> There are new features but there are a lot of BC issues. Maybe this should be targeted for a new version.

Indeed. Should be a new major (e.g. 3.0).

@nlzet no possibility to keep BC with a deprecation system?

nlzet:
@Soullivaneuh no, I don't think it is possible, I tried but I kept going into new BC problems. I think it is wiser to wait for 3.0.

Soullivaneuh:
@nlzet well, we'll ping you when a major branch will be started. :+1: 

Please see: https://github.com/sonata-project/SonataAdminBundle/issues/3053

nlzet:
@rande, @Soullivaneuh I saw there are currently some milestones awaiting for a new 2.4 release for the SonataAdminBundle: https://github.com/sonata-project/SonataAdminBundle/milestones/2.4

I'm unsure but on this issue we said we would wait for version 3.0 on the core bundle, is the 3.0 core version related to the 2.4 release of the admin bundle or will this be in a later stadium ?

Soullivaneuh:
> is the 3.0 core version related to the 2.4 release

No. 3.0 will be shipped when we will have to introduce BC break. Version constraint are not defined in advance AFAIK.

nlzet:
Okay, clear. Just ping me when this day comes.

OskarStark:
Fixes #219

OskarStark:
@nlzet can you please change the docs and add the BC break stuff to the UPGRADE.md?

Thank you very much

nlzet:
@OskarStark I will look into this, probably next week or after. I see that there are already some new versions and features in the eonasdan-datetimepicker, and also there is a work in progress for a new major release. Probably worth looking a little bit deeper to see what is actually going on and what could be useful for Sonata.

Is there a release date target for this PR / the 3.0 milestone?

OskarStark:
glad to hear that @nlzet 

not yet, but we would like to tag the next next MINOR, and after that we will tag the next major too

thank you :+1:

Form/Type/BasePickerType.php
```diff
@@ -85,26 +83,34 @@ public function finishView(FormView $view, FormInterface $form, array $options)
...
+            'dp_use_current'                => true,
+            'dp_min_date'                   => '1/1/1900',
+            'dp_max_date'                   => false,
+            'dp_show_today_button'          => true,
+            'dp_locale'                     => \Locale::getDefault(), // 'en'
+            'dp_default_date'               => false,
+            'dp_disabled_dates'             => false,
+            'dp_enabled_dates'              => false,
+            'dp_icons'                      => array(
+                'time'      => 'glyphicon glyphicon-time',
```
Koc: looks like glyphicon was replaced by font-awesome

Soullivaneuh:
Please rebase and fill the needs tags.

core23:
Please rebase and use the new PR template @nlzet

nlzet:
@Soullivaneuh @core23 sorry for the delay. I'm going to plan some time for this.

I'm wondering, since this is outdated and has been under discussion because of BC breaks. Which branch should I target for this feature ? (Probably need a new PR ?)

core23:
If you can handle this without a BC break, you should target the `3.x` branch, if this can't be done the `master` branch will be your friend.

nlzet:
@core23 at the time it wasn't possible without BC, but I'll have to re-investige the options of the latest version of eonasdan-datepicker, maybe it is possible now. 

Thanks anyway

core23:
Can you have a look at this @nlzet ?

core23:
Please update or close this @nlzet

nlzet:
Sorry @core23,
I closed this, since I cannot find any time to update this.

PR #157, opened 2015-05-14
REMOVE unnecessary transliteration.

chigix:
Signed-off-by: Richard Lea <chigix@zoho.com>

### DIFF
--- a/Twig/Extension/TemplateExtension.php
+++ b/Twig/Extension/TemplateExtension.php
@@ -82,11 +82,6 @@ class TemplateExtension extends \Twig_Extension
         // trim
         $text = trim($text, '-');
 
-        // transliterate
-        if (function_exists('iconv')) {
-            $text = iconv('utf-8', 'us-ascii//TRANSLIT', $text);
-        }
-
         // lowercase
         $text = strtolower($text);
 


chigix:
1. This transliteration would occurs unexcepting notice for some characters like chinese and japanese, which would make twig throw exception.
2. The following includes a `preg_replace` to remove unwanted characters.

Soullivaneuh:
Tests seems to fail.

@rande what do you think of this PR? Don't really know the utility of iconv here.

chigix:
It's my missing on the fix of corresponding code in testing.

Soullivaneuh:
@chigix https://travis-ci.org/sonata-project/SonataCoreBundle/jobs/62518871#L346

Test case should not be adapted IMO, the expected value should not change.

rande:
This is some old legacy code, most sonata's bundles now use cocur's slugify library (but not by default to avoid BC break). So I think the service should be adapted to accept a configuration setting to used cocur's slugify library

OskarStark:
Any news on this PR?

chigix:
Through this disscussion, I'm also very confused that if the slugify provided in SonataCore is to be deprecated? If not, then how to deal with the legacy code which is hard to understand? Sincerely, completing this code in method and testing, rebase commitment is not the main challenge, I think.

OskarStark:
ping @rande

rande:
We need to keep this code as it used in other component. However you can send a new PR based on SonataPageBundle code: https://github.com/sonata-project/SonataPageBundle/blob/master/SonataPageBundle.php#L42-L64

Also this PR need to be rebased.

chigix:
So what should do to deal with the failed testing with the error from the tranliteration? The `iconv` usage previously is not compatible with the characters out of alphabetics, such as Chinese and Japanese, which is the primary focus in this PR. But the remove on this `iconv` code would not be passed through the preset testing code. So I was thinking that testing code should also be modified to adapted this case, right?

Koc:
covered by #214 and can be closed IMHO

OskarStark:
:+1: @Koc

PR #158, opened 2015-05-14
Adding double quote to filename attachment header

pborreli:
As of http://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.5.1 filename should be inside double-quote.


pborreli:
mergeplz

PR #162, opened 2015-06-11
Fix deprecated events

### LINKED ISSUE #161: DEPRECATED - Constants PRE_BIND, BIND and POST_BIND
Hello.

I am updated my project from symfony 2.6 to 2.7 and after update i received this deprecated message:

DEPRECATED - Constants PRE_BIND, BIND and POST_BIND in class Symfony\Component\Form\FormEvents are deprecated since version 2.3 and will be removed in 3.0. Use PRE_SUBMIT, SUBMIT and POST_SUBMIT instead.

Soullivaneuh:
Fixes #161

### DIFF
--- a/Form/EventListener/ResizeFormListener.php
+++ b/Form/EventListener/ResizeFormListener.php
@@ -63,8 +63,8 @@ class ResizeFormListener implements EventSubscriberInterface
     {
         return array(
             FormEvents::PRE_SET_DATA    => 'preSetData',
-            FormEvents::PRE_BIND        => 'preBind',
-            FormEvents::BIND            => 'onBind',
+            FormEvents::PRE_SUBMIT      => 'preBind',
+            FormEvents::SUBMIT          => 'onBind',
         );
     }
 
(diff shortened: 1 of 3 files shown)


Form/EventListener/ResizeFormListener.php
```diff
@@ -63,8 +63,8 @@ public static function getSubscribedEvents()
     {
         return array(
             FormEvents::PRE_SET_DATA    => 'preSetData',
-            FormEvents::PRE_BIND        => 'preBind',
-            FormEvents::BIND            => 'onBind',
+            FormEvents::PRE_SUBMIT      => 'preSubmit',
```
rande: I think I fix those code ... also you need to keep the original method name to avoid BC break

Soullivaneuh: Seems not. :-)

Will take a look for BC.

Soullivaneuh: Is BC really needed? Is this listener extended somewhere?

rande: maybe is some people code, methods are public

Soullivaneuh: Ok I'll do this.

Note: 
https://github.com/sonata-project/SonataCoreBundle/blob/master/Form/EventListener/FixCheckboxDataListener.php#L53

This BC trick will not works because `preSubmit` is called.

Do you want me to fix it?

rande: We can also break the code and just add a note and raise an exception if thoses methods are called.

Soullivaneuh: > just add a note

Agree.

> raise an exception if thoses methods are called.

Well, if people doesn't override subscribed events, old method will be **never** called.

Soullivaneuh:
@rande Added a upgrade note.

Form/EventListener/ResizeFormListener.php
Soullivaneuh: I'll add a quick BC prevention.

PR #165, opened 2015-06-15
Fix : Cannot use object of type [Object] as array in preSubmit Event

wadjeroudi:
Hey,

Since the changes from #162, I have an error when trying to add a new entity from a form with a field of type sonata_type_collection.
The $value is an object in some cases. (when there's at least one entity in the collection).

My form was working before this update.

Example Error: Cannot use object of type AppBundle\Entity\MyEntity as array
in vendor/sonata-project/core-bundle/Form/EventListener/ResizeFormListener.php at line 165

### DIFF
--- a/Form/EventListener/ResizeFormListener.php
+++ b/Form/EventListener/ResizeFormListener.php
@@ -163,7 +163,7 @@ class ResizeFormListener implements EventSubscriberInterface
                 $form->add($name, $this->type, $options);
             }
 
-            if (isset($value['_delete'])) {
+            if (is_array($value) && isset($value['_delete'])) {
                 $this->removed[] = $name;
             }
         }


wadjeroudi:
ping @Soullivaneuh

Soullivaneuh:
I don't think this is related to #162. I didn't make any modification on `preSubmit` method.

Soullivaneuh:
@rande I'm not sure of the issue neither of the proposed fix.

I let you this PR then, if you can take a look. Thanks.

Soullivaneuh:
@wadjeroudi are you using stable or dev version of this bundle? If dev, could you please update it?

This patch could fix your issue: https://github.com/sonata-project/SonataCoreBundle/commit/ebb05829aaa9f0fc0afa904bf6b6441ea34cf638

wadjeroudi:
Thanks. It does fix the issue.

PR #168, opened 2015-06-24
Fix datepicker and datetimepicker (basepicker) locale

monsty:
The actual datepicker and datetimepicker is using "\Locale::getDefault()" which is always returning "en".
So, the html view of the calendar is always in english, even if the default_locale is not english.
Now, the datepicker (js options) is using the real default locale.

### DIFF
--- a/Form/Type/BasePickerType.php
+++ b/Form/Type/BasePickerType.php
@@ -16,6 +16,7 @@ use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\DateTimeType;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\Form\FormView;
+use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * Class BasePickerType (to factorize DatePickerType and DateTimePickerType code.
@@ -31,11 +32,30 @@ abstract class BasePickerType extends AbstractType
     private $formatConverter;
 
     /**
-     * @param MomentFormatConverter $formatConverter
+     * @var TranslatorInterface
      */
-    public function __construct(MomentFormatConverter $formatConverter)
+    protected $translator;
+
+    /**
+     * @var string
+     */
+    protected $locale;
+
+    /**
+     * @param MomentFormatConverter $formatConverter
+     * @param TranslatorInterface   $translator
+     */
+    public function __construct(MomentFormatConverter $formatConverter, TranslatorInterface $translator = null)
     {
         $this->formatConverter = $formatConverter;
+        $this->translator = $translator;
+
+        if (null !== $this->translator) {
+            $this->locale = $this->translator->getLocale();
+        } else {
+            @trigger_error('Initializing '.__CLASS__.' without TranslatorInterface is deprecated since 2.4 and will be remove in 3.0.', E_USER_DEPRECATED);
+            $this->locale = \Locale::getDefault();
+        }
     }
 
     /**
@@ -92,7 +112,7 @@ abstract class BasePickerType extends AbstractType
             'dp_min_date'              => '1/1/1900',
             'dp_max_date'              => null,
             'dp_show_today'            => true,
-            'dp_language'              => \Locale::getDefault(), // 'en'
+            'dp_language'              => $this->locale,
             'dp_default_date'          => '',
             'dp_disabled_dates'        => array(),
             'dp_enabled_dates'         => array(),
--- a/Tests/Form/Type/DatePickerTypeTest.php
+++ b/Tests/Form/Type/DatePickerTypeTest.php
@@ -24,6 +24,13 @@ class DatePickerTypeTest extends \PHPUnit_Framework_TestCase
 {
     public function testGetName()
     {
+        $type = new DatePickerType(new MomentFormatConverter(), $this->getMock('Symfony\Component\Translation\TranslatorInterface'));
+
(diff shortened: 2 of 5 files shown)


Form/Type/BasePickerType.php
```diff
@@ -30,12 +31,15 @@
      */
     private $formatConverter;
 
+    protected $translator;
+
     /**
      * @param MomentFormatConverter $formatConverter
      */
-    public function __construct(MomentFormatConverter $formatConverter)
+    public function __construct(MomentFormatConverter $formatConverter, TranslatorInterface $translator)
```
Soullivaneuh: This is BC break IMO. Could you please make it BC?

monsty:
I made it by setting the translator parameter optionnal, is it ok ?

Form/Type/BasePickerType.php
```diff
@@ -30,12 +31,23 @@
...
      * @param MomentFormatConverter $formatConverter
      */
-    public function __construct(MomentFormatConverter $formatConverter)
+    public function __construct(MomentFormatConverter $formatConverter, TranslatorInterface $translator = null)
     {
         $this->formatConverter = $formatConverter;
+        $this->translator = $translator;
+
+        if ($this->translator !== null) {
+            $this->locale = $this->translator->getLocale();
```
Soullivaneuh: Nice catch.

Null translator argument should be deprecated.

Please notify users with a `trigger_error`

Tests/Form/Type/DatePickerTypeTest.php
```diff
@@ -24,7 +24,7 @@ class DatePickerTypeTest extends \PHPUnit_Framework_TestCase
 {
     public function testGetName()
     {
-        $type = new DatePickerType(new MomentFormatConverter());
+        $type = new DatePickerType(new MomentFormatConverter(), $this->getMock('Symfony\Component\Translation\TranslatorInterface'));
```
Soullivaneuh: A legacy test without parameter should be kept in order to test both options.

Tests/Form/Type/DateTimePickerTypeTest.php
```diff
@@ -24,7 +24,7 @@ class DateTimePickerTypeTest extends \PHPUnit_Framework_TestCase
 {
     public function testGetName()
     {
-        $type = new DateTimePickerType(new MomentFormatConverter());
+        $type = new DateTimePickerType(new MomentFormatConverter(), $this->getMock('Symfony\Component\Translation\TranslatorInterface'));
```
Soullivaneuh: A legacy test without parameter should be kept in order to test both options.

monsty: https://github.com/sonata-project/SonataCoreBundle/commit/2b89a9f33eb9080f063a06dfb6195f466006a4eb : Am I supposed to create a test which is catching the deprecated call ?

Soullivaneuh: Not really.

Make one test with the old constructor signature should just to match the condition statement you introduced.

Call it `testLegacyConstructor` to make phpunit happy with it.

See: https://github.com/symfony/phpunit-bridge

Form/Type/BasePickerType.php
```diff
@@ -31,11 +32,30 @@
...
+        } else {
+            @trigger_error('Initializing '.__CLASS__.' without TranslatorInterface is deprecated. The default locale used by BasePicker is now \'en\'', E_USER_DEPRECATED);
```
Soullivaneuh: "is deprecated since 2.4 and will be remove in 3.0."

> The default locale used by BasePicker is now \'en\''

I do not understand this sentence. I would prefer something like "Give TranslatorInterface on second parameter".

Soullivaneuh:
Needs to be on `2.4`.

cc @rande

monsty:
Is the patch validated ?

OskarStark:
it is, but please fix the CS by `make cs`, sometimes this does not work and you have to fix this manually.

This is the error:
```
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) Form/Type/BasePickerType.php (phpdoc_params)
   2) Tests/Form/Type/BasePickerTypeTest.php (unused_use, braces)
```

Please ping me when you updated the PR and sqaushed your commits.
Please use a "good" commit message

OskarStark:
ping @monsty

OskarStark:
ping again @monsty

Form/Type/BasePickerType.php
```diff
@@ -31,11 +32,30 @@
...
      * @param MomentFormatConverter $formatConverter
+     * @param TranslatorInterface $translator
      */
-    public function __construct(MomentFormatConverter $formatConverter)
+    public function __construct(MomentFormatConverter $formatConverter, TranslatorInterface $translator = null)
     {
         $this->formatConverter = $formatConverter;
+        $this->translator = $translator;
+
+        if ($this->translator !== null) {
```
OskarStark: and please use `null !== $this->translator`

core23: Is there a reason why we should use Yoda condtions? @OskarStark

OskarStark: if is see it i recommend it and use it by myself, but not strict

monsty:
Done ! Sorry for the delay (ping @OskarStark)

OskarStark:
Thank you very much @monsty :+1:

PR #177, opened 2015-08-10
Add Twig extensions as a dependency

tgalopin:
As of 2.7 and https://github.com/symfony/symfony-standard/commit/1272c9fff1c6622c683553b545736eda36914e5a, Symfony does not include by default the twig extensions.

But SonataCoreBundle need them (https://github.com/sonata-project/SonataCoreBundle/blob/master/Resources/config/twig.xml#L12) so I added the Twig extensions as a dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "symfony/validator": "~2.3",
         "symfony/property-access": "~2.3",
         "twig/twig": "~1.16",
+        "twig/extensions": "~1.0",
         "cocur/slugify": "*"
     },
     "require-dev": {


Soullivaneuh:
I'm suprises that tests was not failing before because `twig/extensions` should be included long time ago.

Could you please investigate and add some test of those extensions?

rande:
@Soullivaneuh code coverage is not 100%

Soullivaneuh:
Well, time to fix that. :-)

dbu:
SonataAdminBundle does have the dependency to twig/extensions so its really about the CoreBundle. not sure if the twig services really are needed or if you could just check for the existence and otherwise not register those services, as they seem to be unused in the CoreBundle.

for the no failing tests: aparently twig is never loaded, otherwise we would see an error.

tgalopin:
I don't have time to work on this, so I close it, but I think the problem is still here isn't it?

PR #179, opened 2015-08-14
Use the correct Writer for Excel files

lyoshenka:
Despite being called `XlsWriter`, it does not output Excel-formatted files. Instead it outputs HTML tables. I believe the correct class is `XmlExcelWriter`.

### DIFF
--- a/Exporter/Exporter.php
+++ b/Exporter/Exporter.php
@@ -14,7 +14,7 @@ namespace Sonata\CoreBundle\Exporter;
 use Exporter\Source\SourceIteratorInterface;
 use Exporter\Writer\CsvWriter;
 use Exporter\Writer\JsonWriter;
-use Exporter\Writer\XlsWriter;
+use Exporter\Writer\XmlExcelWriter;
 use Exporter\Writer\XmlWriter;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
@@ -33,7 +33,7 @@ class Exporter
     {
         switch ($format) {
             case 'xls':
-                $writer      = new XlsWriter('php://output');
+                $writer      = new XmlExcelWriter('php://output');
                 $contentType = 'application/vnd.ms-excel';
                 break;
             case 'xml':


OskarStark:
ping @rande

rande:
Depends on which version of office you like to support. XML version if for xlsx format. The html version works with any version of office.

rande:
Office 2016 on mac warns users which 2 modals with the current ``XlsWriter``. Maybe it is the time to change. I will prefer to have a new format in the ``xlsx`` then. So we are BC with the code. This need applied on the SonataAdminBundle too.

Soullivaneuh:
> I will prefer to have a new format in the xlsx then. So we are BC with the code. This need applied on the SonataAdminBundle too.

:+1:

Soullivaneuh:
I put it as major but not sure. ping @sonata-project/contributors

greg0ire:
It's BC with the code. It it breaks something for a user, they will fix it by making a PR, but they will still not change their code, will they? I would tag this as minor.

core23:
Can't we use a config key for this? So the user could choose between XLS and XLSX

Soullivaneuh:
> Can't we use a config key for this? So the user could choose between XLS and XLSX

I was thinking about that to. @lyoshenka could you please take a look?

lyoshenka:
Sorry, it's been a while. What would you like me to look at?

greg0ire:
We would like you to develop a configuration key that would default to the writer for `xlsx`, but would allow to choose the writer for `xls`. Would you be ok to develop that? If yes, do you need some guidance?

lyoshenka:
Yea, I'm happy to do that. What's a configuration key?

greg0ire:
[This](https://symfony.com/doc/2.7/cookbook/bundles/configuration.html#main) is what we are taking about.

Soullivaneuh:
Changing to minor after @greg0ire comment.

Soullivaneuh:
According to the new [Sonata version management](https://github.com/sonata-project/SonataAdminBundle/issues/3053) and next major [release plan](https://github.com/sonata-project/SonataAdminBundle/issues/3731), this project has been refactored regarding branching and versioning.

If you see this message, your PR concerns a patch or a minor release and is not targeting the right branch.

So I'm closing this one, but don't see it as a refusal. If you think your work is still relevant and want to continue, feel free to reopen it on the right branch (e.g. the default one).

Regards.

PR #187, opened 2015-09-16
Rollback catalogue option name

netounet:
As the form type use SonataTranslatableChoiceType as parent the translation catalog parameter must be catalogue and not translation_domain


pulzarraider:
@netounet thank you for reporting the problem. I have fixed it in #191 - the `sonata_type_translatable_choice` is deprecated so we have to replace it with `choice`.

PR #191, opened 2015-09-20
Use choice instead of deprecated sonata_type_translatable_choice in BooleanType

pulzarraider:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #187
| License       | MIT

### DIFF
--- a/Form/Type/BooleanType.php
+++ b/Form/Type/BooleanType.php
@@ -64,7 +64,7 @@ class BooleanType extends AbstractType
      */
     public function getParent()
     {
-        return 'sonata_type_translatable_choice';
+        return 'choice';
     }
 
     /**


Soullivaneuh:
How to be sure that is not BC break?

pulzarraider:
Yes, indeed there is small BC break if user use the `catalogue` option. I will fix it in few hours to avoid this. Thank you for mentioning it.

Soullivaneuh:
Thanks, waiting for that. A deprecation error should be thrown instead.

Soullivaneuh:
And BTW, next time, a deprecation should be put on a new minor version in preference, not a patch.

pulzarraider:
@Soullivaneuh done in  #193. There should be no BC break in BooleanType after merging the PR.

> And BTW, next time, a deprecation should be put on a new minor version in preference, not a patch.

The next release of SonataCoreBundle can be tagged as 2.4.0.

Soullivaneuh:
> No new branch is needed as the code is still compatible.

I disagree. Last stable minor should always have a branch for patch fixes.

pulzarraider:
We are talking here aboult SonataCoreBundle, not SonataAdminBundle. If next minor version will have no BC break, the code will be fully compatible. After releasing the 2.4.0 it will be the next last stable minor. 

Look for example at the [SwiftmailerBundle](https://github.com/symfony/SwiftmailerBundle). There is no  branch called 2.2. or 2.3.

pulzarraider:
The branch should be created before some BC breaks - for exmaple if we will create code for SonataCoreBundle 3.0 that removes deprecated functionality.

Soullivaneuh:
Maybe, but it will be clearer to adopt same branching logic for all sonata projects IMO, including sonata-core.

Please see: https://github.com/sonata-project/SonataAdminBundle/issues/3053

pulzarraider:
My opinion: we should release the minor version more often (nice example is [Twig](https://github.com/twigphp/Twig)), resolve all possible BC breaks and don't create new branches if it's not necessarily. I think it will be waste of resources.

Soullivaneuh:
> we should release the minor version more often

I agree. I know the Twig sample, could be discussed. Maybe you can talk about it on the concerned issue. :+1:

PR #193, opened 2015-09-21
Fix possible BC break in BooleanType

pulzarraider:
Related to #191.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

### DIFF
--- a/Form/Type/BooleanType.php
+++ b/Form/Type/BooleanType.php
@@ -15,6 +15,7 @@ namespace Sonata\CoreBundle\Form\Type;
 use Sonata\CoreBundle\Form\DataTransformer\BooleanTypeToBooleanTransformer;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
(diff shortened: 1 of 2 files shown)


Form/Type/BooleanType.php
```diff
@@ -50,12 +56,26 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
+            'translation_domain' => function (Options $options) {
+                if ($options['catalogue']) {
+                    return $options['catalogue'];
+                }
+
+                return $options['translation_domain'];
+            },
+
+            // uncomment in SonataCoreBundle 3.0
+            // 'translation_domain' => 'SonataCoreBundle',
```
Soullivaneuh: Commented code is note a really good thing IMHO.

pulzarraider: Do you prefer adding this note in the UPGRADE-3.0.md file created in #192?

Soullivaneuh: Just a comment like 'Use directly translation_domain in 3.0' would be sufficient.

pulzarraider: Fixed.

Soullivaneuh:
@pulzarraider Do we have tests for both cases?

pulzarraider:
@Soullivaneuh tests added.

PR #195, opened 2015-10-01
fixed missing annotation and target annotation

Ma27:
these annotations are mandatory when using this constraint as annotation on objects.


Ma27:
any progress here?

rande:
@Ma27 is there a way to unit test this code ?

Koc:
@rande I haven't seen when this kind of changes covered with tests... All of the original Symfonys constraints has annotations like in this PR. Seems ok, can you merge?

OskarStark:
Thank you @Ma27 and tahnk you @Koc for your feedback :+1:

PR #199, opened 2015-10-09
Added new base Controller for some sonata bundles

core23:


Soullivaneuh:
@rande What do you think? 

Related to https://github.com/sonata-project/SonataNewsBundle/pull/257#issuecomment-146780106

Don't know if should be a minor or a patch.

Controller/Controller.php
```diff
@@ -0,0 +1,32 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\Controller as BaseController;
+use Symfony\Component\HttpFoundation\Request;
+
+abstract class Controller extends BaseController
```
pulzarraider: Abstract classes should be prefixed with `Abstract`, see [Symfony naming conventions](http://symfony.com/doc/master/contributing/code/standards.html#naming-conventions).

Soullivaneuh: Indeed, good catch @pulzarraider. :+1:

core23: Thanks for the hint. Renamed the class to ``AbstractController``

core23:
@rande Can you please review?

core23:
As the symfony Controller isn't abstract, should this controller be abstract?

OskarStark:
Yes please



Am Freitag, 16. Oktober 2015 schrieb Christian Gripp :

Soullivaneuh:
ping @rande

rande:
I don't see the need of this controller. Ok you save a few lines on each controller, but you add inheritance complexity. Why not using ``$request = $request ?: $this->getRequest()`` when required ? It will be easier to remove when a major version will be introduce.

Soullivaneuh:
The `resolveRequest` private method was already introduced on admin-bundle. It's just a reuse.

BTW, this abstract controller can be used later for other BC tricks and common tools.

rande:
@Soullivaneuh Yep, but it is in the same controller. I think we can see how we can solve futur problems once we get them. No need to over solve problems that don't exist yet. (again it is my 2 cents :)

core23:
Any other opinions for handling the new ``Request`` parameter in ``Controller``s or should we follow the approach by @rande ?

OskarStark:
I am :-1:  for adding a new BaseController

Soullivaneuh:
> I am :-1:  for adding a new BaseController

Could we know why?

OskarStark:
IMO we just make this turn to stay BC, right?

Now we are creating a new BaseController which we should delete in the next major version, if there are no new methods.
But at this point we have to handle the BC for a non existant class.

And the benefit is nearly zero now, unless we really need a BaseController for new functionality as @rande said.

core23:
As @rande and @OskarStark said, the benefit for this is really low. I'll close this for now.

PR #201, opened 2015-10-25
Set specific cocur/slugify version constraint

florianeckerstorfer:
We will be releasing Version 2.0 of [cocur/slugify](https://github.com/cocur/slugify) in the next few months and since it might break compatibility I suggest setting a more strict version constraint (e.g., `~1.4` for the package.


Soullivaneuh:
Thank you @florianeckerstorfer.

PR #204, opened 2015-11-04
[FIX] Doc

Th3Mouk:
The problem occurs when the form type is used without the locale.
`Moment.js` library is able to manage the current locale but without them the `sonata_type_datetime_picker` field is bugged.

### DIFF
--- a/Resources/doc/reference/form_types.rst
+++ b/Resources/doc/reference/form_types.rst
@@ -315,7 +315,7 @@ In your layout, you'll need to add the assets dependencies (feel free to adapt t
     <head>
         <!-- ... -->
         <script type="text/javascript" src="path_to_jquery.min.js"></script>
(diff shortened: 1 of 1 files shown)


OskarStark:
Looks ok to me 

Ping @Soullivaneuh

Soullivaneuh:
@Th3Mouk Could you provide a how to reproduce case?

Th3Mouk:
@Soullivaneuh Yes i've the problem on Debian 8 servor on php 5.6.7, on `SiteAdmin.php` `enabledTo` field. Without the locales i've invalid parse error.

Soullivaneuh:
I mean, code to reproduce and the exact error. :-)

Th3Mouk:
No need to code just a fresh install of SonataPageBundle

OskarStark:
I can reproduce this error to, thank you @Th3Mouk

PR #220, opened 2015-12-11
prepend config no matter how things are organised

greg0ire:
Mea culpa : I did not understand why there were several config blocks
when writing this piece of code, but now it is clear : in my
application, I am the configuring sonata admin security part in
security.yml, and the other parts in config.yml . The part in security
is the first config block, but it does not contain the options key (the
part in config.yml does). This change will prepend the config as soon as
the requested config key is detected, no matter whether the
configuration is split or not. If this key appears twice in different
config blocks, the last occurrence will have its value taken into
account.

### DIFF
--- a/DependencyInjection/SonataCoreExtension.php
+++ b/DependencyInjection/SonataCoreExtension.php
@@ -34,11 +34,13 @@ class SonataCoreExtension extends Extension implements PrependExtensionInterface
     {
         $configs = $container->getExtensionConfig('sonata_admin');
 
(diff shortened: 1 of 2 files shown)


rande:
@greg0ire Can you add unit test on the prepend function thanks.

greg0ire:
Ok, I'm on it.

greg0ire:
@rande done

OskarStark:
Thank you @greg0ire

PR #222, opened 2015-12-20
FIX: Check if ApiDocBundle is loaded

core23:


OskarStark:
Thank you @core23 

Can you please fix the CS error in a separate PR if you have some time, this would be awesome :+1:

PR #227, opened 2016-02-11
[WIP] Support only active PHP / symfony versions

core23:
Maybe we could use this Travis configuration for all bundles:
 - ~~remove support / allow failures for PHP 5.3 and 5.4~~
 - remove support for outdated symfony version (2.4, 2.5 and 2.6)
 - run **php cs fixer** only once under PHP 7 for performance
 - run  **sphinx-build** only once under PHP 7 for performance

With the next major version for each bundle, we should drop the old php version in composer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,8 @@
 language: php
 
 php:
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
   - hhvm
+  - nightly
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


core23:
@sonata-project/contributors are you okay with this?

OskarStark:
i'm :+1:

core23:
Added spinx-build, inspired by sphinx-build https://github.com/sonata-project/SonataAdminBundle/pull/3355

dbu:
i am not sure about ignoring 5.3 and 5.4 failures. is that already failing now? i totally agree on dropping support for them and outdated symfony versions - that can also be a minor version change, its not a new feature to get rid of old versions. but ignoring failures could bite people - rather update composer json to indicate it simply does not work with those versions. 

imho, increasing the minimum version requirements is less of a BC break than silently start introducing failures with php 5.3...

core23:
Updated composer versions

composer.json
```diff
@@ -16,6 +16,7 @@
         }
     ],
     "require": {
+        "php": "~5.5|~7.0",
         "symfony/http-foundation": "~2.3|~3.0",
```
dbu: this should be `2.3.*|^2.7.0|^3.0.0` i think. currently it allows 2.4 and 2.5 and 2.6 which is too permissive. same for the other symfony dependencies below.

core23: Good point :+1:

Th3Mouk:
@core23 I'm not totally agree with the deletion of 5.3 and 5.4 PHP testing, symfony 2.7 and 2.8 are still `>=5.3.9`, this can injure the stability of bundles.
Concerning the sphinx build i personally don't have it, and I can't run make test command. (I run with phpunit command). I propose to create 2 commands, `make test` and `make build`.

core23:
I know that Symfony does this, but as I said earlier: nobody should use old unsupported versions in a productive environment. @Th3Mouk

Th3Mouk:
I'm ok with the idea, but we need to be aware of current users and current production environnement. I already imagine ton of issues opened, if you decide to drop the support do it on the next major release 3.0.

dbu:
@Th3Mouk dropping support for PHP 5.3 is not a BC break. people will not suddenly get failures after updating on their PHP 5.3 systems. if we have to ignore failures on 5.3 however, and still advertise to support 5.3 in composer.json, people could end up with broken installations after the update.

i vote to up the version requirement in the next minor version. if i am on 5.3, i would simply not get any more updates of sonata. which is not a big deal, given that my underlying PHP is a security risk as well.

Th3Mouk:
:+1:

PR #232, opened 2016-03-10
Modifications in the convert function of the MomentFormatConverter class

ejkun:
As debated in this [issue](https://github.com/sonata-project/SonataAdminBundle/issues/3643). I've changed the convert function of the MomentFormatConverter class, since there was a problem with the output in case of pt_BR locale. I also added a assertion in the MomentFormartConvertTest class, to test the problem.

### DIFF
--- a/Date/MomentFormatConverter.php
+++ b/Date/MomentFormatConverter.php
@@ -46,8 +46,6 @@ class MomentFormatConverter
         'EE' => 'ddd', 'EEEEEE' => 'dd',
         // timezone
         'ZZZZZ' => 'Z', 'ZZZ' => 'ZZ',
-        // letter 'T'
-        '\'T\'' => 'T',
     );
 
     /**
@@ -59,6 +57,48 @@ class MomentFormatConverter
      */
     public function convert($format)
     {
-        return strtr($format, self::$formatConvertRules);
+        $size = strlen($format);
+        $foundOne = false;
+        $output = '';
+        //process the format string letter by letter
+        for ($i = 0; $i < $size; ++$i) {
+            //if finds a '
+            if ($format[$i] == "'") {
+                //if the next character are T' forming 'T', send a T to the
+                //output
+                if ($format[$i + 1] == 'T' && $format[$i + 2] == '\'') {
+                    $output .= 'T';
+                    $i += 2;
+                } else {
+                    //if it's no a 'T' then send whatever is inside the '' to
+                //the output, but send it inside [] (useful for cases like
+                //the brazilian translation that uses a 'de' in the date)
+                    $output .= '[';
+                    $temp = current(explode("'", substr($format, $i + 1)));
+                    $output .= $temp;
+                    $output .= ']';
+                    $i += strlen($temp) + 1;
+                }
+            } else {
+                //if no ' is found, then search all the rules, see if any of
+                //them matchs
+                $foundOne = false;
+                foreach (self::$formatConvertRules as $key => $value) {
+                    if (substr($format, $i, strlen($key)) == $key) {
+                        $output .= $value;
+                        $foundOne = true;
+                        $i += strlen($key) - 1;
+                        break;
+                    }
+                }
+        //if no rule is matched, then just add the character to the
+        //output
+                if (!$foundOne) {
+                    $output .= $format[$i];
+                }
+            }
+        }
+
+        return $output;
     }
 }
--- a/Tests/Date/MomentFormatConverterTest.php
+++ b/Tests/Date/MomentFormatConverterTest.php
@@ -78,5 +78,8 @@ class MomentFormatConverterTest extends \PHPUnit_Framework_TestCase
 
         $phpFormat = 'D MMM y';
         $this->assertEquals('D MMM YYYY', $mfc->convert($phpFormat));
+
+        $phpFormat = "dd 'de' MMMM 'de' YYYY"; //Brazilian date format
+        $this->assertEquals('DD [de] MMMM [de] YYYY', $mfc->convert($phpFormat));
     }
 }


Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
      */
     public function convert($format)
     {
-        return strtr($format, self::$formatConvertRules);
+        $size = strlen($format);
+        $foundOne = false;
+        $output = '';
+        //process the format string letter by letter
+        for ($i = 0; $i < $size; $i++) {
```
greg0ire: Why not use `foreach( str_split($format) as $char)`? You would spare computing `$size`…

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+            //if finds a '
+            if ($format[$i] == '\'') {
```
greg0ire: using `"'"` would be easier to read IMO because there would be no escaping.

Date/MomentFormatConverter.php
greg0ire: Oh I see, you need $i later because it's not letter to letter replacement, some groups of letter are replaced with others.

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                //if the next character are T' forming 'T', send a T to the output
+                if ($format[$i+1] == 'T' && $format[$i+2] == '\'') {
```
greg0ire: Same here, please use `"'"`

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                    $output = $output . 'T';
+                    $i = $i+2;
```
greg0ire: $i += 2

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                } else {
+                    //if it's no a 'T' then send whatever is inside the '' to the output, but send it inside [] (useful for cases like the brazilian translation that uses a 'de' in the date)
+                    $output = $output . '[';
```
greg0ire: And so on

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+        //process the format string letter by letter
+        for ($i = 0; $i < $size; $i++) {
+            //if finds a '
+            if ($format[$i] == '\'') {
+                //if the next character are T' forming 'T', send a T to the output
+                if ($format[$i+1] == 'T' && $format[$i+2] == '\'') {
+                    $output = $output . 'T';
+                    $i = $i+2;
+                } else {
+                    //if it's no a 'T' then send whatever is inside the '' to the output, but send it inside [] (useful for cases like the brazilian translation that uses a 'de' in the date)
```
greg0ire: Please wrap your comments so that the line is not longer than 80 chars (`gqq` in vim)

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                    $output = $output . '[';
+                    $temp = explode('\'', substr($format, $i+1))[0];
+                    $output = $output . $temp;
+                    $output = $output . ']';
+                    $i = $i+strlen($temp)+1;
+                }
+            } else {
+                //if don't find a ' then search all the rules, and see if any on them matchs
```
greg0ire: "if no ' is found" would be better IMO

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                if ($format[$i+1] == 'T' && $format[$i+2] == '\'') {
+                    $output = $output . 'T';
+                    $i = $i+2;
+                } else {
+                    //if it's no a 'T' then send whatever is inside the '' to the output, but send it inside [] (useful for cases like the brazilian translation that uses a 'de' in the date)
+                    $output = $output . '[';
+                    $temp = explode('\'', substr($format, $i+1))[0];
+                    $output = $output . $temp;
+                    $output = $output . ']';
+                    $i = $i+strlen($temp)+1;
```
greg0ire: I think PSR state that you should have spaces before and after the `+` signs

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                foreach (self::$formatConvertRules as $key => $value) {
+                    if (substr($format, $i, strlen($key)) == $key) {
+                        $output = $output . $value;
+                        $foundOne = true;
+                        $i = $i+strlen($key)-1;
+                        break;
+                    }
+                }
+                //if no rule is matched, then just add the character to the output
+                if ($foundOne == false) {
```
greg0ire: `if (!$foundOne)` would be better. Don't compare booleans to booleans. Name them so that it makes a nice sentence when prepended with `if`.

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,42 @@ class MomentFormatConverter
...
+                    $i = $i+strlen($temp)+1;
+                }
+            } else {
+                //if don't find a ' then search all the rules, and see if any on them matchs
+                $foundOne = false;
+                foreach (self::$formatConvertRules as $key => $value) {
+                    if (substr($format, $i, strlen($key)) == $key) {
+                        $output = $output . $value;
+                        $foundOne = true;
+                        $i = $i+strlen($key)-1;
```
greg0ire: Spacing problem

greg0ire:
Just type `make cs`, it should fix the build :)

greg0ire:
Regarding your commit message : it is too hazy. You should put something like this :

```
avoid converting escaped character sequences

Some character sequences that are not meant to be ignored by the date formatter
are escaped with quotes. The intl to moment converter was too naive before and
converted them. This change avoids this conversion, and escapes these character
sequences with brackets, as required by the format used by moment.js
```

greg0ire:
Congratulation on writing this parser :) You did great!

ejkun:
Sorry for the delay, I'm back, did all the changes. Why is the check failing?

greg0ire:
Because the `[` notation for array does not work with old versions of php. You will have to use `array()`

ejkun:
How can I make this check locally?

greg0ire:
I don't think you can if you do not have 5.3 . But I can't see any array creation notation, so it must be something else.

greg0ire:
Link to the error : https://travis-ci.org/sonata-project/SonataCoreBundle/jobs/116119380#L722

Date/MomentFormatConverter.php
```diff
@@ -59,6 +57,48 @@ class MomentFormatConverter
...
+                //output
+                if ($format[$i + 1] == 'T' && $format[$i + 2] == '\'') {
+                    $output .= 'T';
+                    $i += 2;
+                } else {
+                    //if it's no a 'T' then send whatever is inside the '' to
+                //the output, but send it inside [] (useful for cases like
+                //the brazilian translation that uses a 'de' in the date)
+                    $output .= '[';
+                    $temp = explode("'", substr($format, $i + 1))[0];
```
greg0ire: There is your problem. You can't do that in php 5.3

ejkun:
I think the problem is this line
`$temp = explode("'", substr($format, $i + 1))[0];`
I changed it to:
`$temp = current(explode("'", substr($format, $i + 1)));`
I will update my repo and see if it works

Date/MomentFormatConverter.php
greg0ire: But you can use current, like that : `$temp = current(explode("'", substr($format, $i + 1)));`

greg0ire:
Congrats, it's working! The whole build should be green very soon.

greg0ire:
ping @OskarStark @core23 LGTM

core23:
Thank you @ejkun

PR #235, opened 2016-04-08
Return FQCN for SF2.8+ and text type for >2.8

ethernal:
I was editing in the browser - can't do much from work as we are locked behind proxy that's where the errors come from. I did try to squash commits and I can see that it did that locally, but when I synced it with GitHub app I can still see 9 commits to merge. What am I doing wrong

### DIFF
--- a/Form/Type/BooleanType.php
+++ b/Form/Type/BooleanType.php
@@ -81,7 +81,9 @@ class BooleanType extends AbstractType
      */
     public function getParent()
     {
-        return 'choice';
+        return method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix') ?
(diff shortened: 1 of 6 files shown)


ethernal:
@greg0ire sorry for the mess. I was unable to push the squashed commits into the branch, not sure what I was doing wrong. I've redone all of it fresh and now it's in a single commit.
Hope this fixes most deprecation messages now.

greg0ire:
Please run make cs, and commit amend your commit with the result.

greg0ire:
The build also fails because the `getParent()` methods are tested (for the sake of code coverage ?). @core23 @OskarStark @Soullivaneuh what do you think? Should @ethernal change the tests? Or simply remove them?

Soullivaneuh:
:+1: For tests changes.

PR #237, opened 2016-04-22
Return FQCN for SF2.8+ and text type for >2.8

Soullivaneuh:
Closes #235.

Returns fuly qualified class name for the parent in case Symfony 2.8+is
used or text for versions below 2.8


Soullivaneuh:
Thank you @ethernal.

ethernal:
Sorry I was unable to write the tests. I did not really know how for once, and I just tried now but as I can see it's done already, I'll check the code and learn from it for the future.

Thank you @Soullivaneuh  for finishing it.

PR #228, opened 2016-02-11
#issue choice options in advanced filters are flipped [sf3]

rafaelcalleja:
Related to [ sonata-project/SonataAdminBundle@PR #3585](https://github.com/sonata-project/SonataAdminBundle/pull/3585)

Symfony 3.* remove choices_as_value = false 

> symfony/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php

```php
if (true !== $choicesAsValues) {
    throw new \RuntimeException(sprintf('The "choices_as_values" option of the %s should not be used. Remove it and flip the contents of the "choices" option instead.', get_class($this)));
}
```

[source code](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/Extension/Core/Type/ChoiceType.php#L269)

PR #238, opened 2016-04-22
Flip choices depending on the version of Symfony

Soullivaneuh:
Closes #228.

### DIFF
--- a/Form/Type/BooleanType.php
+++ b/Form/Type/BooleanType.php
@@ -60,12 +60,7 @@ class BooleanType extends AbstractType
             self::TYPE_NO  => 'label_type_no',
         );
 
-        if (!method_exists('Symfony\Component\Form\FormTypeInterface', 'setDefaultOptions')) {
-            $choices = array_flip($choices);
-        }
-
-        $resolver->setDefaults(array(
-            'choices'   => $choices,
+        $defaultOptions = array(
             'transform' => false,
 
             // @deprecated Deprecated as of SonataCoreBundle 2.3.10, to be removed in 3.0.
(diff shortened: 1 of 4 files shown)


Form/Type/BooleanType.php
```diff
@@ -79,7 +74,21 @@ public function configureOptions(OptionsResolver $resolver)
...
                 return $options['translation_domain'];
             },
-        ));
+        );
+
+        // SF 2.7+ BC
+        if (method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
+            $choices = array_flip($choices);
+
+            // choice_as_value options is not needed of SF 3.0+
```
greg0ire: "needed as of" ?

Form/Type/EqualType.php
```diff
@@ -56,13 +56,21 @@ public function configureOptions(OptionsResolver $resolver)
             self::TYPE_IS_NOT_EQUAL => $this->translator->trans('label_type_not_equals', array(), 'SonataCoreBundle'),
         );
 
-        if (!method_exists('Symfony\Component\Form\FormTypeInterface', 'setDefaultOptions')) {
+        $defaultOptions = [];
```
greg0ire: Here is what breaks the build.

Form/Type/BooleanType.php
```diff
@@ -73,7 +74,21 @@ public function configureOptions(OptionsResolver $resolver)
 
                 return $options['translation_domain'];
             },
-        ));
+        );
+
+        // SF 2.7+ BC
+        if (method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
+            $choices = array_flip($choices);
```
maxim-mamonov: Looks like this is broke BooleanFilter in the Sonata admin

Soullivaneuh: Could you please open an issue with more details about that?

PR #239, opened 2016-04-22
StyleCI integration

Soullivaneuh:


Command/SonataDumpDoctrineMetaCommand.php
```diff
@@ -1,9 +1,9 @@
 <?php
 
 /*
- * This file is part of the Sonata Connector project.
+ * This file is part of the Sonata Project package.
  *
- * (c) Sylvain Rascar <sylvain.rascar@fullsix.com>
```
OskarStark: Souldnt we add him like an author in this file? @Soullivaneauh

PR #244, opened 2016-05-06
Support form types that don't have a service

WouterJ:
Form types don't have to have a corresponding service tagged as form.type. If they have no dependencies, the FQCN can be instantiated directly.

This is especially important as most core form types are no longer registered as services since Symfony 3.1. Currently, Sonata does not work on Symfony 3.1 because of this.

Btw, I think it would make more sense if this BC layer works the other way around: Using the form name instead of FQCN in Symfony <2.8. This makes the changes required when dropping Symfony <2.8 support way less and makes things as this much easier to handle (as we could just return false from `hasType()` and the FormRegistry would take care of instantiating the class). However, if wanted, let's do that in another PR.

### DIFF
--- a/Form/Extension/DependencyInjectionExtension.php
+++ b/Form/Extension/DependencyInjectionExtension.php
@@ -90,7 +90,11 @@ class DependencyInjectionExtension implements FormExtensionInterface
         $name = self::findClass($this->mappingTypes, $name);
 
         if (!isset($this->typeServiceIds[$name])) {
-            throw new InvalidArgumentException(sprintf('The field type "%s" is not registered with the service container.', $name));
+            if (class_exists($name) && in_array('Symfony\Component\Form\FormTypeInterface', class_implements($name), true)) {
+                return new $name();
+            } else {
+                throw new InvalidArgumentException(sprintf('The field type "%s" is not registered with the service container.', $name));
+            }
         }
 
         $type = $this->container->get($this->typeServiceIds[$name]);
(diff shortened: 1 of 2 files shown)


Form/Extension/DependencyInjectionExtension.php
```diff
@@ -90,7 +90,11 @@ public function getType($name)
         $name = self::findClass($this->mappingTypes, $name);
 
         if (!isset($this->typeServiceIds[$name])) {
-            throw new InvalidArgumentException(sprintf('The field type "%s" is not registered with the service container.', $name));
+            if (class_exists($name) && in_array('Symfony\Component\Form\FormTypeInterface', class_implements($name))) {
```
WouterJ: this duplicates https://github.com/symfony/form/blob/master/FormRegistry.php#L84

Soullivaneuh:
Could you please:

* Add related test?
* Fix StyleCI error?

Thanks.

Soullivaneuh:
BTW, core-bundle 3.0 will be release very soon and the target branch would not be the same.

You will be warned at this time so you will be able to re-open the PR.

WouterJ:
Updated the PR.

Given that this is a bug in a 3.0 feature, I think it would make much sense to include this in 3.0.0.

Soullivaneuh:
Thank you @WouterJ!

WouterJ:
Quick merge, great. Thanks!

PR #270, opened 2016-05-25
Inject writers in the exporter.

### LINKED ISSUE #269: Make the writers configurable services
My client wants to open their CSV files with Microsoft Excel or LibreOffice, but Excel does not automatically detect that commas are in used and defaults to SSV (Semicolon-separated values).

I think a way to fix that would be to make the CsvWriter configurable services (and the other writers while we're at it).

An ugly workaround is to create another Exporter configured with semicolons, and overwrite the service definition, like this : 

```yml
    sonata.admin.exporter: # ugly service id override
        class: MyProject\AppBundle\Admin\Export\Exporter
```

The DIP violation is here : https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Exporter/Exporter.php#L48

greg0ire:
Closes #269 

### Changelog

```markdown
### Added
- `Sonata\CoreBundle\Exporter\Exporter::__construct()` now accepts a hash of writers.
- `Sonata\CoreBundle\Exporter\Exporter::__addWriter()` allows to add writers through the new `sonata_core.exporter.writer` tag.
- Existing exporter writers can be configured with container parameters.
```

### Todo

- [x] test the service definitions
- [x] document the new service
- [x] introduce container parameters to make the default services configurable
- [x] find a way to let people add new writers
- [x] find a way to let people pick and choose default writers

### Subject

Being able to inject any writer means being able to inject custom writers, but also configure existing writers otherwise. Both are interesting.

### DIFF
--- a/DependencyInjection/Compiler/ExporterCompilerPass.php
+++ b/DependencyInjection/Compiler/ExporterCompilerPass.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @author Grégoire Paris <postmaster@greg0ire.fr>
+ */
+final class ExporterCompilerPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has('sonata.core.exporter')) {
+            return;
+        }
+
+        $definition = $container->findDefinition('sonata.core.exporter');
+        $writers = $container->findTaggedServiceIds('sonata.core.exporter.writer');
+
+        foreach (array_keys($writers) as $id) {
+            $definition->addMethodCall(
+                'addWriter',
+                array(new Reference($id))
+            );
+        }
+    }
+}
--- a/DependencyInjection/SonataCoreExtension.php
+++ b/DependencyInjection/SonataCoreExtension.php
@@ -62,7 +62,9 @@ class SonataCoreExtension extends Extension implements PrependExtensionInterface
         $loader->load('twig.xml');
         $loader->load('model_adapter.xml');
         $loader->load('core.xml');
+        $loader->load('exporter.xml');
 
+        $this->configureExporter($container, $config);
         $this->registerFlashTypes($container, $config);
         $container->setParameter('sonata.core.form_type', $config['form_type']);
 
@@ -172,6 +174,17 @@ class SonataCoreExtension extends Extension implements PrependExtensionInterface
 
             $definition = $container->getDefinition('sonata.core.slugify.native');
             $definition->setDeprecated(true);
+        }
+    }
+
+    private function configureExporter(ContainerBuilder $container, array $config)
+    {
+        foreach (array('csv', 'json', 'xls', 'xml') as $format) {
+            if (in_array($format, $config['exporter']['default_writers'])) {
+                $container->getDefinition('sonata.core.exporter.writer.'.$format)->addTag(
+                    'sonata.core.exporter.writer'
+                );
+            }
         }
     }
 }
--- a/Exporter/Exporter.php
+++ b/Exporter/Exporter.php
@@ -14,12 +14,46 @@ namespace Sonata\CoreBundle\Exporter;
 use Exporter\Source\SourceIteratorInterface;
 use Exporter\Writer\CsvWriter;
 use Exporter\Writer\JsonWriter;
+use Exporter\Writer\TypedWriterInterface;
 use Exporter\Writer\XlsWriter;
 use Exporter\Writer\XmlWriter;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
 class Exporter
 {
+    /**
+     * @var TypedWriterInterface[]
+     */
+    private $writers;
+
+    /**
+     * NEXT_MAJOR: change default value to array().
+     *
+     * @param TypedWriterInterface[] $writers an array of allowed typed writers, indexed by format name
+     */
+    public function __construct(array $writers = null)
+    {
+        $this->writers = array();
+        // NEXT_MAJOR: remove this fallback system
+        if ($writers === null) {
+            @trigger_error(
+                'Not supplying writers to the Exporter is deprecated and will not be supported in 4.0.',
+                E_USER_DEPRECATED
+            );
+
+            $this->addWriter(new CsvWriter('php://output', ',', '"', '', true, true));
+            $this->addWriter(new JsonWriter('php://output'));
+            $this->addWriter(new XlsWriter('php://output'));
+            $this->addWriter(new XmlWriter('php://output'));
+
+            return;
+        }
+
+        foreach ($writers as $writer) {
+            $this->addWriter($writer);
+        }
+    }
+
     /**
      * @throws \RuntimeException
      *
@@ -31,35 +65,36 @@ class Exporter
      */
     public function getResponse($format, $filename, SourceIteratorInterface $source)
     {
-        switch ($format) {
-            case 'xls':
-                $writer = new XlsWriter('php://output');
-                $contentType = 'application/vnd.ms-excel';
-                break;
-            case 'xml':
-                $writer = new XmlWriter('php://output');
-                $contentType = 'text/xml';
-                break;
-            case 'json':
-                $writer = new JsonWriter('php://output');
-                $contentType = 'application/json';
-                break;
-            case 'csv':
-                $writer = new CsvWriter('php://output', ',', '"', '', true, true);
-                $contentType = 'text/csv';
-                break;
-            default:
-                throw new \RuntimeException('Invalid format');
+        if (!array_key_exists($format, $this->writers)) {
+            throw new \RuntimeException(sprintf(
+                'Invalid "%s" format, supported formats are : "%s"',
+                $format,
+                implode(', ', array_keys($this->writers))
+            ));
         }
+        $writer = $this->writers[$format];
 
         $callback = function () use ($source, $writer) {
             $handler = \Exporter\Handler::create($source, $writer);
             $handler->export();
         };
 
-        return new StreamedResponse($callback, 200, array(
-            'Content-Type' => $contentType,
+        $headers = array(
             'Content-Disposition' => sprintf('attachment; filename="%s"', $filename),
-        ));
+        );
+
+        $headers['Content-Type'] = $writer->getDefaultMimeType();
+
+        return new StreamedResponse($callback, 200, $headers);
+    }
+
+    /**
+     * The main benefit of this method is the type hinting.
+     *
+     * @param TypedWriterInterface $writer a possible writer for exporting data
+     */
+    private function addWriter(TypedWriterInterface $writer)
+    {
+        $this->writers[$writer->getFormat()] = $writer;
     }
 }
--- a/Resources/config/exporter.xml
+++ b/Resources/config/exporter.xml
@@ -1,6 +1,43 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <parameters>
+        <parameter key="sonata.core.exporter.writer.csv.filename">php://output</parameter>
+        <parameter key="sonata.core.exporter.writer.csv.delimiter">,</parameter>
+        <parameter key="sonata.core.exporter.writer.csv.enclosure">"</parameter>
+        <parameter key="sonata.core.exporter.writer.csv.escape">\</parameter>
+        <parameter key="sonata.core.exporter.writer.csv.show_headers">true</parameter>
+        <parameter key="sonata.core.exporter.writer.csv.with_bom">false</parameter>
+        <parameter key="sonata.core.exporter.writer.json.filename">php://output</parameter>
+        <parameter key="sonata.core.exporter.writer.xls.filename">php://output</parameter>
+        <parameter key="sonata.core.exporter.writer.xls.show_headers">true</parameter>
+        <parameter key="sonata.core.exporter.writer.xml.filename">php://output</parameter>
+        <parameter key="sonata.core.exporter.writer.xml.main_element">datas</parameter>
+        <parameter key="sonata.core.exporter.writer.xml.child_element">data</parameter>
+    </parameters>
     <services>
-        <service id="sonata.core.exporter" class="Sonata\CoreBundle\Exporter\Exporter"/>
+        <service id="sonata.core.exporter.writer.csv" class="Exporter\Writer\CsvWriter" public="false">
+            <argument>%sonata.core.exporter.writer.csv.filename%</argument>
+            <argument>%sonata.core.exporter.writer.csv.delimiter%</argument>
+            <argument>%sonata.core.exporter.writer.csv.enclosure%</argument>
+            <argument>%sonata.core.exporter.writer.csv.escape%</argument>
+            <argument>%sonata.core.exporter.writer.csv.show_headers%</argument>
+            <argument>%sonata.core.exporter.writer.csv.with_bom%</argument>
+        </service>
+        <service id="sonata.core.exporter.writer.json" class="Exporter\Writer\JsonWriter" public="false">
+            <argument>%sonata.core.exporter.writer.json.filename%</argument>
+        </service>
+        <service id="sonata.core.exporter.writer.xls" class="Exporter\Writer\XlsWriter" public="false">
+            <argument>%sonata.core.exporter.writer.xls.filename%</argument>
+            <argument>%sonata.core.exporter.writer.xls.show_headers%</argument>
+        </service>
+        <service id="sonata.core.exporter.writer.xml" class="Exporter\Writer\XmlWriter" public="false">
+            <argument>%sonata.core.exporter.writer.xml.filename%</argument>
+            <argument>%sonata.core.exporter.writer.xml.main_element%</argument>
+            <argument>%sonata.core.exporter.writer.xml.child_element%</argument>
+        </service>
+        <service id="sonata.core.exporter" class="Sonata\CoreBundle\Exporter\Exporter">
+            <argument type="collection"/>
+            <!-- NEXT_MAJOR : remove this argument -->
+        </service>
     </services>
 </container>
--- a/Tests/Exporter/ExporterTest.php
+++ b/Tests/Exporter/ExporterTest.php
@@ -12,19 +12,62 @@
 namespace Sonata\CoreBundle\Tests\Exporter;
 
 use Exporter\Source\ArraySourceIterator;
+use Exporter\Writer\CsvWriter;
+use Exporter\Writer\JsonWriter;
+use Exporter\Writer\XlsWriter;
+use Exporter\Writer\XmlWriter;
 use Sonata\CoreBundle\Exporter\Exporter;
 
 class ExporterTest extends \PHPUnit_Framework_TestCase
 {
-    /**
-     * @expectedException RuntimeException
-     */
     public function testFilter()
     {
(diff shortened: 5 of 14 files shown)


Exporter/Exporter.php
```diff
@@ -14,13 +14,51 @@
...
+    {
+        $this->writers = $writers;
+
+        // NEXT MAJOR : remove this fallback system
+        if ($writers === array()) {
+            $this->writers = array(
+                'csv' => new CsvWriter('php://output', ',', '"', '', true, true),
+                'json' => new JsonWriter('php://output'),
+                'xls' => new XlsWriter('php://output'),
+                'xml' => new XmlWriter('php://output'),
```
greg0ire: Maybe `WriterInterface` shoud have a `getFormatName` method?

Exporter/Exporter.php
```diff
@@ -31,35 +69,39 @@ class Exporter
...
+        $writer = $this->writers[$format];
 
         $callback = function () use ($source, $writer) {
             $handler = \Exporter\Handler::create($source, $writer);
             $handler->export();
         };
 
         return new StreamedResponse($callback, 200, array(
-            'Content-Type' => $contentType,
+            'Content-Type' => $this->getContentTypeForWriter($writer),
```
OskarStark: what about `getContentType($writer)` ?

greg0ire: @OskarStark fixed

Tests/Exporter/ExporterTest.php
```diff
@@ -16,17 +16,21 @@
 
 class ExporterTest extends \PHPUnit_Framework_TestCase
 {
-    /**
-     * @expectedException RuntimeException
-     */
     public function testFilter()
     {
+        $this->setExpectedException('RuntimeException', 'Invalid "foo" format');
```
Soullivaneuh: Ninja hack, this is not related to the PR.

Exporter/Exporter.php
```diff
@@ -14,13 +14,51 @@
...
+ */
 class Exporter
 {
     /**
+     * @var array[string => WriterInterface]
+     */
+    private $writers;
+
+    /**
+     * NEXT MAJOR : remove default value.
```
Soullivaneuh: `NEXT_MAJOR:`

Exporter/Exporter.php
```diff
@@ -14,13 +14,51 @@
 use Exporter\Source\SourceIteratorInterface;
 use Exporter\Writer\CsvWriter;
 use Exporter\Writer\JsonWriter;
+use Exporter\Writer\WriterInterface;
 use Exporter\Writer\XlsWriter;
 use Exporter\Writer\XmlWriter;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
+/**
+ * NEXT MAJOR : make this class final.
```
Soullivaneuh: `NEXT_MAJOR:`

Soullivaneuh: And this is not related to the PR i think. Would be better to have an issue for that.

Exporter/Exporter.php
```diff
@@ -14,13 +14,51 @@
...
+    /**
+     * NEXT MAJOR : remove default value.
+     *
+     * @param array[string => WriterInterface] writers a hash of allowed writers
+     */
+    public function __construct(array $writers = array())
+    {
+        $this->writers = $writers;
+
+        // NEXT MAJOR : remove this fallback system
```
Soullivaneuh: * `NEXT_MAJOR:`
* Is this for BC? Could we put some deprecation warning?

Tests/Exporter/ExporterTest.php
greg0ire: I slightly improved the error message, and wanted to test that. I can make it a separate commit if necessary, and that will be cumbersome, but a separate PR for that would be too much.

Exporter/Exporter.php
greg0ire: Yes it is, I'll add a deprecation warning.

greg0ire: It forced me to rewrite the tests with the new way, which is a good thing :+1:

Tests/Exporter/ExporterTest.php
```diff
@@ -12,21 +12,30 @@
...
         $source = $this->getMock('Exporter\Source\SourceIteratorInterface');
+        $writer = $this->getMock('Exporter\Writer\WriterInterface');
 
-        $exporter = new Exporter();
```
OskarStark: we should keep a test with an empty constructor, tot test if the fallback works, shouldn't we?

greg0ire: You're right, I'll mark it as `@legacy`

greg0ire: @OskarStark done

greg0ire:
Related : https://github.com/sonata-project/exporter/pull/90

greg0ire:

Exporter/Exporter.php
```diff
@@ -31,35 +71,39 @@ class Exporter
...
+    private function getContentType(WriterInterface $writer)
+    {
+        $mapping = array(
+            'Exporter\Writer\CsvWriter' => 'text/csv',
+            'Exporter\Writer\JsonWriter' => 'application/json',
+            'Exporter\Writer\XlsWriter' => 'application/vnd.ms-excel',
+            'Exporter\Writer\XmlWriter' => 'text/xml',
+        );
+
+        return $mapping[get_class($writer)];
```
OskarStark: this could result in an undefined index, couldn't it?

greg0ire: Uuuuh you're right! How did I not see this?!

greg0ire: @OskarStark fixed

Exporter/Exporter.php
```diff
@@ -14,13 +14,53 @@
...
 {
     /**
+     * @var array[string => WriterInterface]
+     */
+    private $writers;
+
+    /**
+     * NEXT MAJOR: remove default value.
+     *
+     * @param array[string => WriterInterface] writers a hash of allowed writers
```
core23: Should be `@param WriterInterace[] ...`

AFAIK your notation is not offical supported and doesn't work with autocompletion.

greg0ire: Do you mean it prevents autocompletion ?

greg0ire: And then, there is this other notation : http://stackoverflow.com/a/1763425/353612

greg0ire: So it should be `WriterInterface[]` ?

greg0ire: @core23 fixed

OskarStark:
LGTM 👍 

@Soullivaneuh can you please give us a final review?

greg0ire:
Travis says "offline for maintenance" :S

Tests/Exporter/ExporterTest.php
```diff
@@ -12,21 +12,61 @@
...
     }
 
+    public function testConstructorRejectsNonWriters()
+    {
+        $this->setExpectedException('InvalidArgumentException', 'Writer for format "made-up-format"');
+        new Exporter(array('made-up-format' => 'Not even an object'));
+    }
+
+    /**
+     * NEXT MAJOR: remove this test.
```
Soullivaneuh: `NEXT_MAJOR`

https://github.com/sonata-project/SonataCoreBundle/blob/3.x/CONTRIBUTING.md#the-base-branch

Exporter/Exporter.php
```diff
@@ -31,35 +71,39 @@ class Exporter
...
 
-        return new StreamedResponse($callback, 200, array(
-            'Content-Type' => $contentType,
+        $headers = array(
             'Content-Disposition' => sprintf('attachment; filename="%s"', $filename),
-        ));
+        );
+
+        /*
+         * NEXT MAJOR: rely on https://github.com/sonata-project/exporter/pull/90
```
Soullivaneuh: Why should this be done on next major?

Soullivaneuh:
@greg0ire Should we wait https://github.com/sonata-project/exporter/pull/90#issuecomment-222369290?

Exporter/Exporter.php
greg0ire: Because it will imply changing the composer version constraints on the exporter, which is a BC-break

Tests/Exporter/ExporterTest.php
greg0ire: @Soullivaneuh fixed

greg0ire:
@Soullivaneuh : why is StyleCI failing ?

greg0ire:
> @greg0ire Should we wait sonata-project/exporter#90 (comment)?

Yes, it might be better, if it is rejected or amended for some reason.

Exporter/Exporter.php
Soullivaneuh: Your PR is for a new minor version of exporter. Changing version constraint for a new minor is not BC break.

Exporter/Exporter.php
```diff
@@ -31,35 +71,39 @@ class Exporter
...
 
-        return new StreamedResponse($callback, 200, array(
-            'Content-Type' => $contentType,
+        $headers = array(
             'Content-Disposition' => sprintf('attachment; filename="%s"', $filename),
-        ));
+        );
+
+        /*
+         * NEXT_MAJOR: rely on https://github.com/sonata-project/exporter/pull/90
```
Soullivaneuh: Not sure about that, why on next major?

The `getMimeType` returns in https://github.com/sonata-project/exporter/pull/90 is the same, isn't it?

greg0ire: Because it implies bumping the version constraint for exporter

Soullivaneuh: Yes, but a minor one, right?

In this case, this is not BC break.

Sullivan SENECHAL

greg0ire: > In this case, this is not BC break.

as seen yesterday on twitter you mean ? Right…

greg0ire: Well let's wait for sonata-project/exporter#90 then, shall we ?

Soullivaneuh: > as seen yesterday on twitter you mean ? Right…

No. Yesterday I was talking about PHP. You know, the gray area...

Here, this is a constraint change for a minor version of a composer package.

Plus, it's a package of ours and we have sure the BC way is kept.

So no, it's not BC break at all.

The same way was adopt for PR concerning class renaming from `Admin` to `AbstractAdmin`

> Well let's wait for sonata-project/exporter#90 then, shall we ?

Yes. :+1:

OskarStark:
@greg0ire what is the status of this PR?

greg0ire:
Blocked by sonata-project/exporter#90, which is blocked by https://github.com/sonata-project/exporter/pull/114 . Please unblock the whole chain guys.

greg0ire:
More precisely : this need to be reworked once sonata-project/exporter#90 is merged, because it will be greatly simplified.

Soullivaneuh:
> More precisely : this need to be reworked once sonata-project/exporter#90 is merged, because it will be greatly simplified.

And I think wait for a new release of exporter, am I right?

greg0ire:
> And I think wait for a new release of exporter, am I right?

Yes, exactly. This MR should touch the Composer manifest. Which by our standards, is a BC-break. Which means I should reopen it on master, closing it for now.

Soullivaneuh:
> Which by our standards, is a BC-break.

Only for important changes like PHP or Symfony.

For the other, except we have to change for a next major of a deps, this is not needed.

Plus, it's an internal project.

greg0ire:
Ok then :)

Soullivaneuh:
@greg0ire You can now play with it: https://github.com/sonata-project/exporter/releases/tag/1.5.0

greg0ire:
Great, will do!

Exporter/Exporter.php
```diff
@@ -14,13 +14,45 @@
...
+     * @var TypedWriterInterface[]
+     */
+    private $writers;
+
+    /**
+     * NEXT_MAJOR: remove default value.
+     *
+     * @param TypedWriterInterface[] $writers an array of allowed typed writers, indexed by format name
+     */
+    public function __construct(array $writers = array())
```
greg0ire: It would be a BC-break, no ?

greg0ire: ping @Soullivaneuh

Soullivaneuh: Oh yes, you are right.

Tests/Exporter/ExporterTest.php
```diff
@@ -12,21 +12,64 @@
...
 
-        $exporter = new Exporter();
+        $exporter = new Exporter(array($writer));
         $exporter->getResponse('foo', 'foo', $source);
     }
 
+    public function testConstructorRejectsNonTypedWriters()
+    {
+        $this->setExpectedException(
+            PHP_VERSION_ID < 70000 ? 'PHPUnit_Framework_Error' : 'TypeError',
```
greg0ire: @Soullivaneuh : I tried detecting `TypeError` here instead, but there is a polyfill for it, which means it will not work.

greg0ire:
@sonata-project/contributors , this is ready for review.

Tests/Exporter/ExporterTest.php
Soullivaneuh: First of all, why trying to expect a phpunit exception?

greg0ire: It's a feature of phpunit, it converts errors to exception, so that you can assert that an error should occur. But do you mean I should except something less precise, like maybe `\Exception` ?

Soullivaneuh: Oh ok, special feature for PHP < 7.0?

Ok I think this is acceptable. But please use `version_compare`, this is more readable.

greg0ire: @Soullivaneuh done

Exporter/Exporter.php
```diff
@@ -31,35 +63,36 @@ class Exporter
...
+    }
+
+    /**
+     * The main benefit of this method is the type hinting.
+     *
+     * @param TypedWriterInterface $writer a possible writer for exporting data
+     */
+    final private function addWriter(TypedWriterInterface $writer)
+    {
+        $this->writers[$writer->getFormat()] = $writer;
```
Soullivaneuh: Why not just get an associative array on the constructor?

Soullivaneuh:
`Exporter` is a service. Nothing to update about this?

Soullivaneuh:
**PLEASE DON'T MERGE IT FOR NOW**

I found a nasty issue about exporter service. I'll dig on it.

Soullivaneuh:
Wait https://github.com/sonata-project/SonataCoreBundle/pull/308 and a patch release before merge.

Exporter/Exporter.php
greg0ire: Because that way you benefit from the type hinting (that's what I tried to explain in the method's documentation)

Soullivaneuh:
@greg0ire Could you please rebase and update your PR?

greg0ire:
@Soullivaneuh done

greg0ire:
But it needs some tests, and I think I will add container parameters for the csv writer in this PR (or do you feel otherwise?)

Resources/config/exporter.xml
```diff
@@ -1,6 +1,30 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <service id="sonata.core.exporter" class="Sonata\CoreBundle\Exporter\Exporter"/>
+        <service id="sonata.core.exporter.writer.csv" class="Exporter\Writer\CsvWriter" public="false">
+            <argument>php://output</argument>
+            <argument>,</argument>
```
Soullivaneuh: Hmm does not `CsvWriter` have default values for that? Maybe same for the first argument.

BTW, you made the services private. But what if the user want to change this?

Resources/config/exporter.xml
```diff
@@ -1,6 +1,30 @@
...
+        </service>
+        <service id="sonata.core.exporter.writer.xls" class="Exporter\Writer\XlsWriter" public="false">
+            <argument>php://output</argument>
+        </service>
+        <service id="sonata.core.exporter.writer.xml" class="Exporter\Writer\XmlWriter" public="false">
+            <argument>php://output</argument>
+        </service>
+        <service id="sonata.core.exporter" class="Sonata\CoreBundle\Exporter\Exporter">
+            <argument type="collection">
+                <argument type="service" id="sonata.core.exporter.writer.csv" />
```
Soullivaneuh: Same here. You pass an array of service. Good.

But if I want to add one writer on it without dropping the default ones?

This is maybe for this @rande talked about factories...

Resources/config/exporter.xml
greg0ire: This is still in progress. I'll keep them private and use container parameters instead

Resources/config/exporter.xml
greg0ire: I think this should be solved with tagged services

greg0ire: Also, maybe it is already possible to add a method call to `addWriter` (not sure if this works for private services)

greg0ire:
I ended a with the following : 

1. If the exporter is called without parameters, the `$writers` defaults to null, and this is the legacy case, to be removed on next major.
2. If the exporter is called with an array, add all the writers in the array
3. To avoid the deprecation notice introduced by 1 but still be BC, I inject an empty array, and then, in the extension, I tag services that are part of the list of default writers.

OskarStark:
regarding the CHANGELOG:

what about:
```diff
- The constructor method of `Sonata\CoreBundle\Exporter\Exporter` now accepts a hash of writers
+ `Sonata\CoreBundle\Exporter\Exporter::__construct()` now accepts a hash of writers
```

greg0ire:
Good idea, I'll use that instead!

DependencyInjection/Compiler/ExporterCompilerPass.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Class StatusRendererCompilerPass.
+ *
+ *
+ * @author Grégoire Paris <postmaster@greg0ire.fr>
+ */
+class ExporterCompilerPass implements CompilerPassInterface
```
core23: Final?

DependencyInjection/Compiler/ExporterCompilerPass.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace Sonata\CoreBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Class StatusRendererCompilerPass.
```
core23: We don't need this

Exporter/Exporter.php
```diff
@@ -31,35 +65,36 @@ class Exporter
...
+
+        return new StreamedResponse($callback, 200, $headers);
+    }
+
+    /**
+     * The main benefit of this method is the type hinting.
+     *
+     * @param TypedWriterInterface $writer a possible writer for exporting data
+     */
+    final private function addWriter(TypedWriterInterface $writer)
```
core23: Why final? Nobody can extend this method.

Soullivaneuh: On some tricky way, they actually can.

core23: Started a new discussion to keep this PR clean.

https://github.com/sonata-project/dev-kit/issues/185

DependencyInjection/Compiler/ExporterCompilerPass.php
greg0ire: Plus, it's wrong :P

Exporter/Exporter.php
greg0ire: Fixed

DependencyInjection/SonataCoreExtension.php
```diff
@@ -174,4 +176,15 @@ protected function deprecateSlugify(ContainerBuilder $container)
             $definition->setDeprecated(true);
         }
     }
+
+    final private function configureExporter(ContainerBuilder $container, array $config)
```
core23: -final

greg0ire: Done, sorry… :sweat_smile:

OskarStark:
LGTM 👍

greg0ire:
ping @sonata-project/contributors : do we merge this here first, then move it to the exporter library, or should I do this inside the library directly?

Soullivaneuh:
If we move it to exporter after, this mean deprecated fresh functionalities, right?

In this case, the second option looks better.

greg0ire:
> If we move it to exporter after, this mean deprecated fresh functionalities, right?

Right. I'll rework this exporter-side.

PR #288, opened 2016-06-01
use new valid composer syntax

OskarStark:
```markdown
### Changed
- Use new composer syntax in `composer.json`
```

### Subject

The build was failing, due to a new composer lint check

### Example

https://travis-ci.org/sonata-project/SonataCoreBundle/jobs/134350181

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,13 +27,13 @@
         "cocur/slugify": "^1.4 || ^2.0"
     },
     "require-dev": {
-        "matthiasnoback/symfony-config-test": "~0.4 || ^1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "~0.7",
+        "matthiasnoback/symfony-config-test": "^0.4 || ^1.0",
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7",
         "doctrine/orm": "^2.4",
         "doctrine/phpcr-odm": "^1.0",
         "jackalope/jackalope-doctrine-dbal": "^1.0",
         "friendsofsymfony/rest-bundle": "^1.1 || ^2.0",
-        "jms/serializer-bundle": "~0.11 || ^1.0",
+        "jms/serializer-bundle": "0.11 - 0.13 || ^1.0",
         "sensio/framework-extra-bundle": "^2.3 || ^3.0",
         "sonata-project/exporter": "^1.3",
         "symfony/phpunit-bridge": "^2.7",


OskarStark:
@Soullivaneuh if this is fine, i can check the other repositories, just give me a ping, a second one is created here: https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/577

composer.json
```diff
@@ -27,13 +27,13 @@
...
-        "matthiasnoback/symfony-config-test": "~0.4 || ^1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "~0.7",
+        "matthiasnoback/symfony-config-test": "^0.4 || ^1.0",
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7",
         "doctrine/orm": "^2.4",
         "doctrine/phpcr-odm": "^1.0",
         "jackalope/jackalope-doctrine-dbal": "^1.0",
         "friendsofsymfony/rest-bundle": "^1.1 || ^2.0",
-        "jms/serializer-bundle": "~0.11 || ^1.0",
+        "jms/serializer-bundle": "^0.11 || ^1.0",
```
Soullivaneuh: `^0.13`

Or if you want to keep `~0.11`: `^0.11 || ^0.12 || ^0.13`.

Soullivaneuh:
@OskarStark Yes, thank you! :+1: 

Be careful with `^` and 0.x versions. If you add `^0.11` for example, only `0.11.*` versions would be valid.

This is because a 0.x minor version **can** contains BC break things according to semver.

composer.json
OskarStark: > Or if you want to keep ~0.11: ^0.11 || ^0.12 || ^0.13.

to be honest i don't know what i want here, except a valid syntax 😄

OskarStark: if we would use `^0.11 || ^0.12 || ^0.13` did we run all the tests in all combinations?

Soullivaneuh: > if we would use `^0.11 || ^0.12 || ^0.13` did we run all the tests in all combinations?

What do you mean?

BTW, you can also use `0.11 - 0.13 || ^1.0` here.

OskarStark: > BTW, you can also use 0.11 - 0.13 || ^1.0 here.

nice 👍

OskarStark: >> if we would use ^0.11 || ^0.12 || ^0.13 did we run all the tests in all combinations?
>
> What do you mean?

I mean does travis run all possible combinations or do we trust the maintainer? i mean, how to be sure, that `0.12` is working ?

OskarStark:
Used this: `0.11 - 0.13 || ^1.0`

greg0ire:
Thanks @OskarStark

PR #294, opened 2016-06-07
added test for InlineValidator

ahmetakbn:
### Subject

Test to increase code coverage of `SonataCoreBundle`

### DIFF
--- a/Tests/Fixtures/Bundle/Validator/FooValidatorService.php
+++ b/Tests/Fixtures/Bundle/Validator/FooValidatorService.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Tests\Fixtures\Bundle\Validator;
+
+use Sonata\CoreBundle\Validator\ErrorElement;
+use Symfony\Component\Validator\Exception\ValidatorException;
+
+/**
+ * Validator service to create exception for test.
+ *
+ * @author Ahmet Akbana <ahmetakbana@gmail.com>
+ */
+final class FooValidatorService
+{
+    /**
+     * @param ErrorElement $errorElement
+     * @param string $value
+     * @throws ValidatorException
+     */
+    public function fooValidatorMethod(ErrorElement $errorElement, $value)
+    {
+        throw new ValidatorException($errorElement->getSubject().' is equal to '.$value);
+    }
+}
--- a/Tests/Validator/InlineValidatorTest.php
+++ b/Tests/Validator/InlineValidatorTest.php
@@ -1,0 +1,142 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Tests\Validator;
+
+use Sonata\CoreBundle\Tests\Fixtures\Bundle\Validator\FooValidatorService;
+use Sonata\CoreBundle\Validator\ErrorElement;
+use Sonata\CoreBundle\Validator\InlineValidator;
+use Symfony\Component\Validator\Exception\ValidatorException;
+
+/**
+ * @author Ahmet Akbana <ahmetakbana@gmail.com>
+ */
+final class InlineValidatorTest extends \PHPUnit_Framework_TestCase
+{
+    private $container;
+    private $constraintValidatorFactory;
+    private $context;
+
+    public function setUp()
+    {
+        $this->container = $this->getMock('Symfony\Component\DependencyInjection\ContainerInterface');
+        $this->constraintValidatorFactory = $this->getMock(
+            'Symfony\Component\Validator\ConstraintValidatorFactoryInterface'
+        );
+        $this->context = $this->getMock(
+            interface_exists('Symfony\Component\Validator\Context\ExecutionContextInterface') ?
+                'Symfony\Component\Validator\Context\ExecutionContextInterface' :
+                'Symfony\Component\Validator\ExecutionContextInterface'
+        );
+    }
+
+    public function testGetErrorElement()
+    {
+        $inlineValidator = new InlineValidator($this->container, $this->constraintValidatorFactory);
+
+        $inlineValidator->initialize($this->context);
+
+        $reflectorObject = new \ReflectionObject($inlineValidator);
+        $reflectedMethod = $reflectorObject->getMethod('getErrorElement');
+        $reflectedMethod->setAccessible(true);
+
+        $errorElement = $reflectedMethod->invokeArgs($inlineValidator, array('foo'));
+
+        $this->assertInstanceOf('Sonata\CoreBundle\Validator\ErrorElement', $errorElement);
+        $this->assertSame('foo', $errorElement->getSubject());
+    }
+
+    public function testValidateWithConstraintIsClosure()
+    {
+        $this->setExpectedException('Symfony\Component\Validator\Exception\ValidatorException', 'foo is equal to foo');
+
+        $constraint = $this->getMock('Symfony\Component\Validator\Constraint', array('isClosure', 'getClosure'));
+
+        $constraint->expects($this->once())
+            ->method('isClosure')
+            ->willReturn(true);
+
+        $constraint->expects($this->once())
+            ->method('getClosure')
+            ->willReturn(function (ErrorElement $errorElement, $value) {
+                throw new ValidatorException($errorElement->getSubject().' is equal to '.$value);
+            });
+
+        $inlineValidator = new InlineValidator($this->container, $this->constraintValidatorFactory);
+
+        $inlineValidator->initialize($this->context);
+
+        $inlineValidator->validate('foo', $constraint);
+    }
+
+    public function testValidateWithConstraintGetServiceIsString()
+    {
+        $this->setExpectedException('Symfony\Component\Validator\Exception\ValidatorException', 'foo is equal to foo');
+
+        $constraint = $this->getMock('Symfony\Component\Validator\Constraint', array(
+            'isClosure',
+            'getService',
+            'getMethod',
+        ));
+
+        $constraint->expects($this->once())
+            ->method('isClosure')
+            ->willReturn(false);
+
+        $constraint->expects($this->any())
+            ->method('getService')
+            ->willReturn('string');
+
+        $constraint->expects($this->once())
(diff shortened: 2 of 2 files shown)


Tests/Validator/InlineValidatorTest.php
```diff
@@ -0,0 +1,162 @@
...
+        $this->context = $this->getMock(
+            interface_exists('Symfony\Component\Validator\Context\ExecutionContextInterface') ?
+                'Symfony\Component\Validator\Context\ExecutionContextInterface' :
+                'Symfony\Component\Validator\ExecutionContextInterface'
+        );
+    }
+
+    public function testGetErrorElement()
+    {
+        $inlineValidator = new InlineValidator($this->container, $this->constraintValidatorFactory);
```
greg0ire: Mother of god… do all classes depend on the container or what? T___T

Tests/Validator/InlineValidatorTest.php
```diff
@@ -0,0 +1,162 @@
...
+        $reflectedMethod->setAccessible(true);
+
+        $errorElement = $reflectedMethod->invokeArgs($inlineValidator, array('foo'));
+
+        $this->assertInstanceOf('Sonata\CoreBundle\Validator\ErrorElement', $errorElement);
+        $this->assertSame('foo', $errorElement->getSubject());
+    }
+
+    /**
+     * @expectedException \Exception
```
OskarStark: please use `$this->setExpectedException` instead of the annoations

Tests/Validator/InlineValidatorTest.php
OskarStark: > Mother of god… do all classes depend on the container or what? T___T

😆

greg0ire: I think this should have been done with tagged services or something like that…

Tests/Validator/InlineValidatorTest.php
core23: Nope, `setExpectedException` is deprecated, since phpunit 5.2.8

OskarStark: really?

Tests/Validator/InlineValidatorTest.php
OskarStark: but please lets do this in another PR

Tests/Validator/InlineValidatorTest.php
OskarStark: so we keep it this way?

Tests/Validator/InlineValidatorTest.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    public function testGetErrorElement()
+    {
+        $inlineValidator = new InlineValidator($this->container, $this->constraintValidatorFactory);
+
+        $reflectorClass = new \ReflectionClass($inlineValidator);
+        $reflectedProperty = $reflectorClass->getProperty('context');
+        $reflectedProperty->setAccessible(true);
+        $reflectedProperty->setValue($inlineValidator, $this->context);
```
greg0ire: Why not use `initialize` here ?

Tests/Validator/InlineValidatorTest.php
core23: You shouldn't expect the god `\Exception` class

Tests/Validator/InlineValidatorTest.php
```diff
@@ -0,0 +1,162 @@
...
+            'Symfony\Component\Validator\ConstraintValidatorFactoryInterface'
+        );
+        $this->context = $this->getMock(
+            interface_exists('Symfony\Component\Validator\Context\ExecutionContextInterface') ?
+                'Symfony\Component\Validator\Context\ExecutionContextInterface' :
+                'Symfony\Component\Validator\ExecutionContextInterface'
+        );
+    }
+
+    public function testGetErrorElement()
```
greg0ire: Not a huge fan of testing protected methods… you should only test them trough the public API of the object IMO : we don't really care about the implementation, only about the behavior of the system under test. If it is not simple to test like though, it might be justified…

Tests/Validator/InlineValidatorTest.php
Soullivaneuh: > so we keep it this way?

No. Please use this method.

The method is deprecated, not removed.

We will change it when we can drop PHPUnit 4 usage.

greg0ire: Related : https://github.com/sebastianbergmann/phpunit/issues/2074

Tests/Fixtures/Bundle/Validator/FooValidatorService.php
```diff
@@ -0,0 +1,32 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Tests\Fixtures\Bundle\Validator;
+
+use Sonata\CoreBundle\Validator\ErrorElement;
+
+/**
+ * Class FooValidatorService.
```
Soullivaneuh: Useless comment.

Tests/Validator/InlineValidatorTest.php
greg0ire: It's a bit tricky : it is deprecated, but not in favor of annotations. It's deprecated in favor of several methods that are not available to us : `expectException`, etc…

Tests/Validator/InlineValidatorTest.php
greg0ire: > but please lets do this in another PR

Sure goes without saying, this is more a call to action…

Tests/Validator/InlineValidatorTest.php
Soullivaneuh: > It's a bit tricky : it is deprecated, but not in favor of annotations. It's deprecated in favor of several methods that are not available to us : `expectException`, etc…

So just use methods which are both available on PHPUnit 4 and 5. :-)

ahmetakbn: So I will keep using annotations?

greg0ire: Uuuuh… nope!

Tests/Validator/InlineValidatorTest.php
ahmetakbn: :dizzy_face: ah
Yes of course. I didn't really analyze the parent class, just didn't see any setter for the property.

Tests/Validator/InlineValidatorTest.php
ahmetakbn: First of all, just trying to write test for every line which is possible. Form my point of view the protected methods are not %100 encapsulated, so testing them when it is possible/easy is always beneficial.

Tests/Fixtures/Bundle/Validator/FooValidatorService.php
ahmetakbn: This has been created automaticaly by the IDE. I will change it ok.

Tests/Validator/InlineValidatorTest.php
greg0ire: Do as you see fit, just making sure you know my point of view. If it is really a hassle to test the protected method through the public methods, then by all means, do the reflection dark magic (just my opinion though, not sure what @sonata-project/contributors think about this).

Tests/Fixtures/Bundle/Validator/FooValidatorService.php
Soullivaneuh: I know I think we have the same. :+1: 

You can change it on settings.

Tests/Validator/InlineValidatorTest.php
ahmetakbn: I didn't really understand what I really should use finally :smile:

greg0ire: You're supposed to use `$this->setExpectedException('Some\PreciseException', 'some exception message')`

Tests/Validator/InlineValidatorTest.php
ahmetakbn: done.

Tests/Validator/InlineValidatorTest.php
ahmetakbn: done

Tests/Fixtures/Bundle/Validator/FooValidatorService.php
ahmetakbn: I have changed the comment.

Tests/Validator/InlineValidatorTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+        $this->assertInstanceOf('Sonata\CoreBundle\Validator\ErrorElement', $errorElement);
+        $this->assertSame('foo', $errorElement->getSubject());
+    }
+
+    public function testValidateWithConstraintIsClosure()
+    {
+        $this->setExpectedException('Symfony\Component\Validator\Exception\ValidatorException', 'foo is equal to foo');
+
+        $constraint = $this->getMock('Symfony\Component\Validator\Constraint', array('isClosure', 'getClosure'));
```
greg0ire: `getMock` is deprecated in favor of `createMock`

Soullivaneuh: Yes, but `createMock` is not available under PHPUnit 4. So let it as is for now.

greg0ire: Oh, right… nevermind, then.

ahmetakbn:
This PR is ready from my side

Tests/Fixtures/Bundle/Validator/FooValidatorService.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Sonata\CoreBundle\Validator\ErrorElement;
+use Symfony\Component\Validator\Exception\ValidatorException;
+
+/**
+ * Validator service to create exception for test.
+ *
+ * @author Ahmet Akbana <ahmetakbana@gmail.com>
+ */
+class FooValidatorService
```
Soullivaneuh: `final`

ahmetakbn: done.

Tests/Validator/InlineValidatorTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+namespace Sonata\CoreBundle\Tests\Validator;
+
+use Sonata\CoreBundle\Tests\Fixtures\Bundle\Validator\FooValidatorService;
+use Sonata\CoreBundle\Validator\ErrorElement;
+use Sonata\CoreBundle\Validator\InlineValidator;
+use Symfony\Component\Validator\Exception\ValidatorException;
+
+/**
+ * Test for InlineValidator.
```
Soullivaneuh: For me this comment is useless.

By `class InlineValidatorTest`, we know this class is to test `InlineValidator`. :wink:

ahmetakbn: done

Soullivaneuh:
RTM. Thank you @ahmetakbn! :+1:

OskarStark:
Good job @ahmetakbn !

PR #303, opened 2016-06-15
added test for BaseSerializerHandler

ahmetakbn:
### Subject

Improving tests for `SonataCoreBundle`

### DIFF
--- a/Tests/Serializer/BaseSerializerHandlerTest.php
+++ b/Tests/Serializer/BaseSerializerHandlerTest.php
@@ -1,0 +1,144 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Tests\Serializer;
+
+use JMS\Serializer\GraphNavigator;
+use Sonata\CoreBundle\Tests\Fixtures\Bundle\Serializer\FooSerializer;
+
+/**
+ * @author Ahmet Akbana <ahmetakbana@gmail.com>
+ */
+final class BaseSerializerHandlerTest extends \PHPUnit_Framework_TestCase
+{
+    public function testGetSubscribingMethods()
+    {
+        $manager = $this->getMock('Sonata\CoreBundle\Model\ManagerInterface');
+
+        $serializer = new FooSerializer($manager);
+
+        $expectedMethods = array(
+            array(
+                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
+                'format' => 'json',
+                'type' => 'foo',
+                'method' => 'serializeObjectToId',
+            ),
+            array(
+                'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
+                'format' => 'json',
+                'type' => 'foo',
+                'method' => 'deserializeObjectFromId',
+            ),
+            array(
+                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
+                'format' => 'xml',
+                'type' => 'foo',
+                'method' => 'serializeObjectToId',
+            ),
+            array(
+                'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
+                'format' => 'xml',
+                'type' => 'foo',
+                'method' => 'deserializeObjectFromId',
+            ),
+            array(
+                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
+                'format' => 'yml',
+                'type' => 'foo',
+                'method' => 'serializeObjectToId',
+            ),
+            array(
+                'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
+                'format' => 'yml',
+                'type' => 'foo',
+                'method' => 'deserializeObjectFromId',
+            ),
+        );
+
+        $methods = $serializer->getSubscribingMethods();
+
+        $this->assertSame($methods, $expectedMethods);
+    }
+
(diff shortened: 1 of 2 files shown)


Tests/Serializer/BaseSerializerHandlerTest.php
```diff
@@ -0,0 +1,144 @@
...
+                'direction' => GraphNavigator::DIRECTION_DESERIALIZATION,
+                'format' => 'yml',
+                'type' => 'foo',
+                'method' => 'deserializeObjectFromId',
+            ),
+        );
+
+        $methods = $serializer->getSubscribingMethods();
+
+        $this->assertSame($methods, $expectedMethods);
```
greg0ire: Doesn't this test feel weird? If yes, it is probably because the formats are hardcoded in the class. Improve that if you feel up to it (by introducing a `formats` property that you might change from your tests), and it the end, the tests should look like this : 
- check that there are twice as many methods as the number of formats
- check that half for them are about deserialization…

Tests/Serializer/BaseSerializerHandlerTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+        $visitor = $this->getMock('JMS\Serializer\VisitorInterface');
+        $visitor->expects($this->once())
+            ->method('visitInteger')
+            ->with(1, array('foo'), $context)
+            ->willReturn(true);
+
+        $serializer = new FooSerializer($manager);
+
+        $this->assertTrue($serializer->serializeObjectToId($visitor, $data, array('foo'), $context));
```
greg0ire: How does this test check that we have an instance of manager ?

Tests/Serializer/BaseSerializerHandlerTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+        $context = $this->getMock('JMS\Serializer\Context');
+
+        $visitor = $this->getMock('JMS\Serializer\VisitorInterface');
+        $visitor->expects($this->never())
+            ->method('visitInteger');
+
+        $serializer = new FooSerializer($manager);
+
+        $serializer->serializeObjectToId($visitor, $data, array('foo'), $context);
```
greg0ire: There is a lot of repetition here. Could you please use a data provider or a private method ?

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: While writing the test for this class, I was aware that the formats were hardcoded. It is indeed weird. But since I was writing test, this was another concern. I don't want to change the structure of the class while writing test for it. This should be business of another PR IMO. After adding the property the tests also must be updated as the new structure requires.

greg0ire: > if you feel up to it

Sure, I understand perfectly, no problem :)

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: I am manuplating the return for `getClass` method of `$manager` to an instance of my fake class `FooSerializer`. And I set the variable `$data` to instance of my fake class as well. 'Data is Instance of Manager' is the logic in the method but test does not work with exact instance of manager of course.

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: What is the repetition exactly I didn't get it. The mocks are a bit different then the mock in previous test method.

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: So there is one more thing to do :)

Tests/Serializer/BaseSerializerHandlerTest.php
```diff
@@ -0,0 +1,144 @@
...
+        );
+
+        $methods = $serializer->getSubscribingMethods();
+
+        $this->assertSame($methods, $expectedMethods);
+    }
+
+    public function testSerializeObjectToIdWithDataIsInstanceOfManager()
+    {
+        $data = $this->getMock(
```
greg0ire: Please find a more descriptive name than `$data`

ahmetakbn: Sure, I can. But this is the name of the original method's parameter.

Tests/Serializer/BaseSerializerHandlerTest.php
greg0ire: Indeed, I read a bit too fast :)

Tests/Serializer/BaseSerializerHandlerTest.php
greg0ire: Ok

Tests/Serializer/BaseSerializerHandlerTest.php
greg0ire: Oh god…

greg0ire: I'd say rename it anyway. If someone wants to change that in the class someday, they probably won't think of looking at the tests too…

Soullivaneuh:
@greg0ire I let you handle this PR.

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: ok, I changed the name to `$modelInstance`

Tests/Serializer/BaseSerializerHandlerTest.php
greg0ire: See #306

greg0ire:
Thanks a lot @ahmetakbn !

PR #308, opened 2016-06-17
Add missing exporter service

Soullivaneuh:
I am targetting this branch, because this is a fix.

## Changelog

```markdown
### Fixed
- Add missing exporter service
```

## Subject

The Exporter class come from AdminBundle project where this is deprecated.

The old Exporter class from AdminBundle has an active service: sonata.admin.exporter

This one does not have any and still rely on the old one, by inheritance.

Adding the sonata.core.exporter service will permit to:

* Update AdminBundle to use the new service and not the deprecated class
* Add a deprecation warning on the old Exporter class without have this one yelling all time on user projects
* Remove the deprecated Exporter class AND the related admin service on next major of admin-bundle

### DIFF
--- a/Resources/config/exporter.xml
+++ b/Resources/config/exporter.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="sonata.core.exporter" class="Sonata\CoreBundle\Exporter\Exporter"/>
+    </services>
+</container>


Soullivaneuh:
ping @sonata-project/contributors

greg0ire:
Makes more sense… BTW, not sure the exporter should be in the `SonataCoreBundle`. It could be in a `Bundle` directory in the exporter lib. Not saying we should not merge your PR, just thinking about the future.

Soullivaneuh:
Totally agree with you.

We can do a bridge part like done here: https://github.com/nexylan/paybox-direct/tree/master/src/Bridge

greg0ire:
> We can do a bridge part like done here: https://github.com/nexylan/paybox-direct/tree/master/src/Bridge

Exactly what I was thinking about, makes more sense.

Soullivaneuh:
@greg0ire Can I let you create an issue on exporter?

rande:
@greg0ire It was done to avoid too many bundles definition ;) As this logic can be "exporter" to other component.

Soullivaneuh:
> @greg0ire It was done to avoid too many bundles definition ;)

This is why we want to add it on exporter as a bridge instead of create yet another bundle. :wink: 

Ref: https://github.com/sonata-project/SonataCoreBundle/pull/308#issuecomment-226751374

greg0ire:
See https://github.com/sonata-project/exporter/issues/118

greg0ire:
Added a PR to deprecate the class (I can't deprecate the service, because the `deprecated` keyword is not available for old versions of symfony). See #328

PR #309, opened 2016-06-17
Added LegacyFormHelper to detect symfony versions

core23:
I am targetting this branch, because this adds a new BC feature.

## Changelog

```markdown
### Added
- Added new `LegacyFormHelper` to handle the different form behavior
```

## Subject

Adds a new useful helper class to detect the symfony versions and handles the different behavior of forms.

### DIFF
--- a/Form/FormHelper.php
+++ b/Form/FormHelper.php
@@ -11,6 +11,7 @@
 
 namespace Sonata\CoreBundle\Form;
 
+use Sonata\CoreBundle\Util\LegacyFormHelper;
 use Symfony\Component\Form\Form;
 use Symfony\Component\Form\FormTypeInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
@@ -93,9 +94,9 @@ class FormHelper
      */
     public static function configureOptions(FormTypeInterface $type, OptionsResolver $optionsResolver)
     {
-        if (!method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+        if (LegacyFormHelper::isLegacy()) {
             $type->setDefaultOptions($optionsResolver);
-        } else { // SF <2.8 BC
+        } else {
             $type->configureOptions($optionsResolver);
         }
     }
--- a/Form/Type/BooleanType.php
+++ b/Form/Type/BooleanType.php
@@ -12,6 +12,7 @@
 namespace Sonata\CoreBundle\Form\Type;
 
 use Sonata\CoreBundle\Form\DataTransformer\BooleanTypeToBooleanTransformer;
+use Sonata\CoreBundle\Util\LegacyFormHelper;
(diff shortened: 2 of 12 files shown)


Koc:
similar to https://github.com/symfony-cmf/routing-bundle/pull/325

Soullivaneuh:
Quite :-1: about that.

This introduce a new class that will have to be deprecated when we will drop Symfony 2.3 support and removed on next major.

In a nutshell: Having to make BC trick on a class handling BC tricks. Meh. :-/

core23:
Maybe adding a deprecation notice, that you shouldn't rely on it when using symfony 2.8+?

Soullivaneuh:
> Maybe adding a deprecation notice, that you shouldn't rely on it when using symfony 2.8+?

Would produce the same result: Deprecated warnings for "nothing".

core23:
But copy & paste code in every class isn't a good idea either.

ksom:
What do you think about having LegacyFormHelper in each sonata bundle which need it. But tag it `@internal` instead of `@deprecated`. Since it's internal there is no need to deprecate it.

Soullivaneuh:
> What do you think about having LegacyFormHelper in each sonata bundle which need it.

So you mean copy paste exact code, right.

What about... NO ? :trollface: 

cc @greg0ire

Soullivaneuh:
> But copy & paste code in every class isn't a good idea either.

In this case, this is a normal function call to handle a BC trick.

If create a class to handle BC things like this was a good idea, Symfony would already made it...

core23:
> This introduces a new class that will have to be deprecated when we will drop Symfony 2.3 support and removed on next major.

IMO that's nothing against this class. We must remove some BC `method_exists`-hacks too when dropping symfony support. This class would help to find most BC hacks.

We should mark this as `@internal` with a soft deprecation notice for symfony 2.3.

greg0ire:
Related discussion : https://github.com/symfony/symfony-docs/issues/6078

Soullivaneuh:
> Related discussion : symfony/symfony-docs#6078

I think stof said something about that too, but can't remember...

Soullivaneuh:
> We should mark this as @internal with a soft deprecation notice for symfony 2.3.

Marking this class `@internal` means "Do not use it outside!".

Using it on another sonata bundle **is** using it outside.

Soullivaneuh:
Plus, you made a `LegacyFormHelper::isLegacy()` method.

Is legacy regarding what? Regarding which version? What if we have to handle another legacy stuff on SF 3.2? This is not clear at all.

greg0ire:
> I think stof said something about that too, but can't remember...

I had an occasion to speak with him about it, and it is a bad idea for him precisely because of the "you'll end up maintaining BC for a lib about BC", so… Compaception.

PR #311, opened 2016-06-19
Add missing dependency

greg0ire:
## Changelog

``` markdown
### Fixed
- A dependency was explicitely required.
```
## Subject

This bundle was missing a dependency, this PR sort packages first, then adds the dependency.

### DIFF
--- a/Twig/Extension/DeprecatedTextExtension.php
+++ b/Twig/Extension/DeprecatedTextExtension.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Twig\Extension;
+
+/**
+ * NEXT_MAJOR : remove this class and the twig/extensions dependency.
+ *
+ * @deprecated since version 3.x, to be removed in 4.0.
+ */
+final class DeprecatedTextExtension extends \Twig_Extensions_Extension_Text
+{
+    public function twig_truncate_filter(Twig_Environment $env, $value, $length = 30, $preserve = false, $separator = '...')
+    {
+        $this->notifyDeprecation();
+
+        return parent::twig_truncate_filter($env, $value, $length, $preserve, $separator);
+    }
+
+    public function twig_wordwrap_filter(Twig_Environment $env, $value, $length = 80, $separator = "\n", $preserve = false)
+    {
+        $this->notifyDeprecation();
+
+        return parent::twig_wordwrap_filter($env, $value, $length, $separator, $preserve);
+    }
+
+    private function notifyDeprecation()
+    {
+        @trigger_error(
+            'Using the sonata.core.twig.extension.text service is deprecated since 3.x and will be removed in 4.0'
+        );
+    }
+}
--- a/Resources/config/twig.xml
+++ b/Resources/config/twig.xml
@@ -5,7 +5,8 @@
             <tag name="twig.extension"/>
             <argument>%sonata.core.form_type%</argument>
         </service>
-        <service id="sonata.core.twig.extension.text" class="Twig_Extensions_Extension_Text">
+        <!-- NEXT_MAJOR : remove this service -->
+        <service id="sonata.core.twig.extension.text" class="Sonata\CoreBundle\Twig\Extension\DeprecatedExtensionText">
             <tag name="twig.extension"/>
         </service>
         <service id="sonata.core.twig.status_extension" class="Sonata\CoreBundle\Twig\Extension\StatusExtension">
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "symfony/translation": "^2.3 || ^3.0",
         "symfony/twig-bridge": "^2.3.5 || ^3.0",
         "symfony/validator": "^2.3 || ^3.0",
+        "twig/extensions": "^1.0",
         "twig/twig": "^1.23"
     },
     "require-dev": {


greg0ire:
To be reviewed commit by commit.

greg0ire:
Here is the external package reference : https://github.com/sonata-project/SonataCoreBundle/blob/e42a9439f4ad9694a8d52c76f92a4be19be4cfad/Resources/config/twig.xml#L8

greg0ire:
ping @sonata-project/contributors

Soullivaneuh:
> Here is the external package reference : https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Resources/config/twig.xml#L8

Sorry but... the hell is this service?! Oo

greg0ire:
See #33

Soullivaneuh:
Well, this should not the role of admin-bundle to load the twig-extension, but the role of the final project.

Am I wrong?

cc @rande @eko

Soullivaneuh:
@greg0ire Waiting the answer, could you please extract the sort package commit on another PR to get this merged?

greg0ire:
> Am I wrong?

Nope

> @greg0ire Waiting the answer, could you please extract the sort package commit on another PR to get this merged?

Will do, I did it like that because I thought it was gonna be a no-brainer :P looks like I was wrong.

Soullivaneuh:
@eko @rande Could we have your though for this?

eko:
Hi @Soullivaneuh,

I can't really remember why this Twig `Text` extension was needed in the `SonataCoreBundle` but I think this was because of the `ecommerce` repository that uses some `|truncate` filters (for instance: https://github.com/sonata-project/ecommerce/blob/50a5ec86e2e65a700c601cc31e25a9378407b735/src/ProductBundle/Resources/skeleton/product/Resources/views/Entity/form_basket_element.html.twig#L23)

Maybe this can be moved to `ecommerce` if it is not needed by the core?

For information, this Twig extension only provides the 2 following filters (see: http://twig.sensiolabs.org/doc/extensions/text.html):
* truncate
* wordwrap

greg0ire:
Thanks for your answer @eko ! Should we merge this nontheless, and remove the dependency in a non-BC PR ?

Soullivaneuh:
> Maybe this can be moved to `ecommerce` if it is not needed by the core?

Yes, but in any case, I don't see the benefit to declare the class as a service.

Why not just using the TwigBundle?

greg0ire:
I think the goal is to avoid to have to do this : http://blog.christian-baer.com/index.php/truncate-text-in-twig-symfony-2/

Soullivaneuh:
Okay so some points to tell:

* We can't remove this service on a patch, neither a major. Lot of people might be using `truncate` thinking this is auto-provided by the twig bundle. But it is not, it is provided by this one. Remove this will produce a lot of in-comprehensive BC break. We can't deprecate the class but we can deprecate the service starting 4.0, when we will bumping SF version to 2.8.
* I still think declaring a service for a vendor class is **not our job**. Instead, we should think why this extension is not loaded automatically and find a proper solution. Maybe we could:
  * Propose to add it as default on symfony/twig-bundle? It's already done for debug: https://github.com/symfony/twig-bundle/blob/master/Resources/config/twig.xml#L122
  * Find and use a bundle dedicated to provide those extension as services?
  * Create an external bundle for that?

Maybe I'm wrong and in this case I'll be glad to hear your argument. But the actual implementation looks very wrong to me.

Soullivaneuh:
For the external bundle, it already exists: https://packagist.org/packages/dms/twig-extension-bundle

The only thing I don't really like is the configuration part: https://github.com/rdohms/dms-twig-extension-bundle#usage

If the end user decide to disable the text part, this will break our bundles.

If we do our bundle, we can do something that will provide service dynamically only if the one is not provided.

Soullivaneuh:
Asked the question on Symfony: https://github.com/symfony/symfony/issues/19130

greg0ire:
I think we should consider getting rid of `truncate` , the usage is very low : 

https://github.com/sonata-project/ecommerce/search?utf8=%E2%9C%93&q=truncate

Soullivaneuh:
Waiting symfony/symfony#19130, we can:

* Deprecate/Remove the extension service from core-bundle
* To avoid future conflict with Twig 2.0, maybe create your own `sonata_truncate` filter on `ecommerce`? This is quite ugly, but I don't have any another idea ATM...

greg0ire:
> To avoid future conflict with Twig 2.0, maybe create your own sonata_truncate filter on ecommerce? This is quite ugly, but I don't have any another idea ATM...

I'd say simply drop it (and do it in CSS instead, as suggested by @iltar)

core23:
Any updates here?

We need this dependency for the current stable and drop it with the next major?

Soullivaneuh:
@greg0ire Let's deprecate it in order to drop it later so.

greg0ire:
How / what would you deprecate? This filter is not part of the project, we can't deprecate it.

Soullivaneuh:
Oh right. Maybe deprecate the service?

greg0ire:
Uuuuh… what service?

Soullivaneuh:
https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Resources/config/twig.xml#L8

greg0ire:
Oh right the service is ours, didn't realize that.

core23:
Any updates here @greg0ire ?

greg0ire:
Forgot that one. I'll try to have a look at it soon.

greg0ire:
> Oh right. Maybe deprecate the service?

The problem is that deprecating a service is not available on old versions of symfony :(
I think I'll introduce an instantly deprecated class inheriting from `Twig_Extensions_Extension_Text` and use it to define the service instead.

greg0ire:
>  I think I'll introduce an instantly deprecated class inheriting from Twig_Extensions_Extension_Text and use it to define the service instead.

Oh but I can't… extending a Twig extension is deprecated IIRC.

greg0ire:
> Oh but I can't… extending a Twig extension is deprecated IIRC.

I'm not so sure about that. Maybe what's deprecated is defining a new class tagged as a Twig extension that inherits from another class also tagged as a twig extension.

Twig/Extension/DeprecatedTextExtension.php
```diff
@@ -0,0 +1,39 @@
...
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Twig\Extension;
+
+/**
+ * NEXT_MAJOR : remove this class and the twig/extensions dependency.
```
core23: Please add a `@deprecated` annotation

greg0ire: done

core23: Think you could remove the `NEXT_MAJOR` comment and use the deprecated annotation:

`@deprecated since version 3.x, to be removed in 4.0.`

greg0ire: source? I think we already discussed this somewhere, but I'm not sure what the outcome was…

core23: Just started a search through the code ;) Haven't found any usage for `NEXT_MAJOR` and `@deprecated`.

Don't have the ticket where we discussed this.

greg0ire: I don't like being unsure, so let's discuss this here : https://github.com/sonata-project/dev-kit/issues/206

OskarStark:
@greg0ire can you finish this PR?
@core23 If everything is fine now, please approve this PR 👍

greg0ire:
I think it is finished…

Twig/Extension/DeprecatedTextExtension.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @deprecated
```
core23: `@deprecated since version 3.x, to be removed in 4.0.`

greg0ire: done

OskarStark:
Thank you @greg0ire 👍

PR #312, opened 2016-06-20
added static setFormats and addFormat methods

### LINKED ISSUE #306: Inject formats in BaseSerializerHandler
### Environment

Not relevant

## Subject

The BaseSerializerHandler has [formats hardcoded in it](https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Serializer/BaseSerializerHandler.php#L43). This should be changed with injection through a method with the following signature : `public static function setFormats`. The injection should be optional and default to the current values (json, xml, and yml).

The issue was discovered [here](https://github.com/sonata-project/SonataCoreBundle/pull/303#discussion_r67187167), thanks to @ahmetakbn new tests, which should be simplified in the same PR.

ahmetakbn:
Closes #306

## Changelog

```markdown

### Added
- Added static `setFormats` and `addFormat` methods to `BaseSerializerHandler`
- Added `sonata.core.serializer.formats` configuration
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -34,6 +34,7 @@ class Configuration implements ConfigurationInterface
         $rootNode = $treeBuilder->root('sonata_core');
 
         $this->addFlashMessageSection($rootNode);
+        $this->addSerializerFormats($rootNode);
 
         $rootNode
             ->children()
@@ -102,6 +103,29 @@ class Configuration implements ConfigurationInterface
                                     ->end()
                                 ->end()
                             ->end()
+                        ->end()
+                    ->end()
+                ->end()
+            ->end()
+        ;
+    }
+
+    /**
+     * Returns configuration for serializer formats.
+     *
+     * @param ArrayNodeDefinition $node
+     */
+    private function addSerializerFormats(ArrayNodeDefinition $node)
+    {
+        $node
+            ->children()
+                ->arrayNode('serializer')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('formats')
+                            ->prototype('scalar')->end()
+                            ->defaultValue(array('json', 'xml', 'yml'))
+                            ->info('Default serializer formats, will be used while getting subscribing methods.')
                         ->end()
                     ->end()
                 ->end()
--- a/DependencyInjection/SonataCoreExtension.php
+++ b/DependencyInjection/SonataCoreExtension.php
@@ -12,6 +12,7 @@
 namespace Sonata\CoreBundle\DependencyInjection;
 
 use Sonata\CoreBundle\Form\FormHelper;
+use Sonata\CoreBundle\Serializer\BaseSerializerHandler;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -69,6 +70,8 @@ class SonataCoreExtension extends Extension implements PrependExtensionInterface
         $this->configureClassesToCompile();
 
         $this->deprecateSlugify($container);
+
+        $this->configureSerializerFormats($config);
     }
 
     public function configureClassesToCompile()
@@ -151,6 +154,14 @@ class SonataCoreExtension extends Extension implements PrependExtensionInterface
         $definition->replaceArgument(3, $cssClasses);
 
         $container->setDefinition($identifier, $definition);
+    }
+
+    /**
+     * @param array $config
+     */
+    public function configureSerializerFormats($config)
+    {
+        BaseSerializerHandler::setFormats($config['serializer']['formats']);
     }
 
     protected function deprecateSlugify(ContainerBuilder $container)
--- a/Resources/doc/reference/serialization.rst
+++ b/Resources/doc/reference/serialization.rst
@@ -7,7 +7,22 @@ Serialization
 Custom handlers
 ---------------
 
-The bundle comes with a ``BaseSerializerHandler`` to let you customize your serialized entities; this handler is used to serialize/deserialize an entity to/from its id, but you remain free to create your own handler for your specific needs.
+The bundle comes with a ``BaseSerializerHandler`` to let you customize your serialized entities;
+this handler is used to serialize/deserialize an entity to/from its id within the defaults
+formats ('json', 'xml', 'yml').
+
+The serializer default formats are configurable. You can change them from the configuration file.
+
+.. code-block:: yaml
+
+        sonata_core:
+            serializer:
+                formats: ['json', 'xml', 'yml']
+
+You can set these formats to a different array or you can add another format to these formats by using
+``BaseSerializerHandler`` methods ``setFormats`` and ``addFormat``
+
+You are free to create your own handler for your specific needs.
 
 Just override ``Sonata\CoreBundle\Serializer\BaseSerializerHandler`` to create a `JMS Serializer` handler.
 
--- a/Serializer/BaseSerializerHandler.php
+++ b/Serializer/BaseSerializerHandler.php
@@ -27,6 +27,11 @@ abstract class BaseSerializerHandler implements SerializerHandlerInterface
     protected $manager;
 
     /**
+     * @var string[]
+     */
+    protected static $formats;
+
+    /**
      * @param ManagerInterface $manager
      */
     public function __construct(ManagerInterface $manager)
@@ -35,15 +40,41 @@ abstract class BaseSerializerHandler implements SerializerHandlerInterface
     }
 
     /**
+     * @param string[] $formats
+     */
+    final public static function setFormats(array $formats)
+    {
+        static::$formats = $formats;
+    }
+
+    /**
+     * @param string $format
+     */
+    final public static function addFormat($format)
+    {
+        static::$formats[] = $format;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public static function getSubscribingMethods()
     {
+        // NEXT_MAJOR : remove this block
+        if (null === static::$formats) {
+            static::$formats = array('json', 'xml', 'yml');
+            @trigger_error(
+                '$formats has been set to default array("json", "xml", "yml"). Setting $formats to a 
+                default array is deprecated since version 3.0 and will be removed in 4.0. Use SonataCoreBundle 
+                configuration to add default serializer formats.',
+                E_USER_DEPRECATED
+            );
+        }
+
         $type = static::getType();
-        $formats = array('json', 'xml', 'yml');
         $methods = array();
 
-        foreach ($formats as $format) {
+        foreach (static::$formats as $format) {
             $methods[] = array(
                 'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
                 'format' => $format,
--- a/Tests/Serializer/BaseSerializerHandlerTest.php
+++ b/Tests/Serializer/BaseSerializerHandlerTest.php
@@ -19,7 +19,12 @@ use Sonata\CoreBundle\Tests\Fixtures\Bundle\Serializer\FooSerializer;
  */
 final class BaseSerializerHandlerTest extends \PHPUnit_Framework_TestCase
 {
-    public function testGetSubscribingMethods()
+    /**
+     * @group legacy
+     *
+     * NEXT_MAJOR : this should call setFormats method
+     */
+    public function testGetSubscribingMethodsWithDefaultFormats()
     {
         $manager = $this->getMock('Sonata\CoreBundle\Model\ManagerInterface');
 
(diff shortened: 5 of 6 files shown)


Serializer/BaseSerializerHandler.php
```diff
@@ -35,15 +40,22 @@ public function __construct(ManagerInterface $manager)
     }
 
     /**
+     * @param array $formats
+     */
+    public static function setFormats(array $formats)
```
Soullivaneuh: * Why static?
* This should be `final`
* Maybe having a `addFormat` would be great too.

Soullivaneuh:
Maybe some documentation could be added? cc @greg0ire

Serializer/BaseSerializerHandler.php
```diff
@@ -27,6 +27,11 @@
     protected $manager;
 
     /**
+     * @var array
```
Soullivaneuh: `string[]`

Serializer/BaseSerializerHandler.php
```diff
@@ -35,15 +40,22 @@ public function __construct(ManagerInterface $manager)
     }
 
     /**
+     * @param array $formats
```
Soullivaneuh: `string[]`

Serializer/BaseSerializerHandler.php
```diff
@@ -35,15 +40,22 @@ public function __construct(ManagerInterface $manager)
...
      * {@inheritdoc}
      */
     public static function getSubscribingMethods()
     {
         $type = static::getType();
-        $formats = array('json', 'xml', 'yml');
         $methods = array();
 
-        foreach ($formats as $format) {
+        foreach (self::$formats as $format) {
```
Soullivaneuh: `static::`

Serializer/BaseSerializerHandler.php
```diff
@@ -27,6 +27,11 @@
     protected $manager;
 
     /**
+     * @var array
+     */
+    private static $formats = array('json', 'xml', 'yml');
```
greg0ire: They are still harcoded, but you can also inject them. Either change your commit message to reflect that, or find a right place to make a call to `setFormats`… but maybe all this would be easier if this was not static ?

Tests/Serializer/BaseSerializerHandlerTest.php
```diff
@@ -69,6 +69,34 @@ public function testGetSubscribingMethods()
         $this->assertSame($methods, $expectedMethods);
     }
 
+    public function testSetFormats()
```
greg0ire: I think `testGetSubscribingMethods` could be removed.

Serializer/BaseSerializerHandler.php
greg0ire: It is static because the `$formats` array is static :(

Serializer/BaseSerializerHandler.php
Soullivaneuh: > but maybe all this would be easier if this was not static ?

I don't think we can. The method using this property is static too.

greg0ire: > The method using this property is static too.

I know…  there is the problem, maybe this should be changed too. IMO static is very often bad design because it prevents proper injection.

greg0ire: You're right, it cannot be changed, because it is requested by JMS serializer : http://jmsyst.com/libs/serializer/master/handlers

ahmetakbn: It has to be hardcoded with the injection option. Otherwise it will be a BC break. I can set this default array in constructor also.

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: IMO this test is testing the default value of array, may be I can change the name of method instead of deleting the whole method.

Serializer/BaseSerializerHandler.php
greg0ire: Yeah but can you be sure the constructor is called before the call to the static method happens ?

greg0ire: >  Either change your commit message to reflect that, or find a right place to make a call to setFormats

Maybe the right place would be the SonataCoreExtension class ? What do you think @Soullivaneuh ?

Tests/Serializer/BaseSerializerHandlerTest.php
greg0ire: It's good to check you did not break anything, so maybe let's indeed keep it, with a different name. Since the values are still hardcoded, it makes sense.

Serializer/BaseSerializerHandler.php
ahmetakbn: for this I need to change the private `formats` property to protected.

Serializer/BaseSerializerHandler.php
ahmetakbn: > Yeah but can you be sure the constructor is called before the call to the static method happens ?

Upps, of course not. Actually that was the reason I set the array when defining the property. Then I was confused when you told somewhere else :)

greg0ire:
What is this ` taqwq ` you are adding ?

Resources/doc/reference/serialization.rst
```diff
@@ -7,7 +7,12 @@ Serialization
 Custom handlers
 ---------------
 
-The bundle comes with a ``BaseSerializerHandler`` to let you customize your serialized entities; this handler is used to serialize/deserialize an entity to/from its id, but you remain free to create your own handler for your specific needs.
+The bundle comes with a ``BaseSerializerHandler`` to let you customize your serialized entities;
+this handler is used to serialize/deserialize an entity to/from its id within the defaults
+formats ('json', 'xml', 'yml'). You can set this formats to a different array or
```
greg0ire: this => these

Resources/doc/reference/serialization.rst
```diff
@@ -44,3 +49,4 @@ And your handler need to specify the type name:
                 return 'post_type';
             }
         }
+
```
greg0ire: Please don't add this line

Serializer/BaseSerializerHandler.php
```diff
@@ -27,6 +27,11 @@
     protected $manager;
 
     /**
+     * @var string[]
+     */
+    protected static $formats = array('json', 'xml', 'yml');
```
greg0ire: It would be great to make this configurable, so

1. this should be `null`
2. if it is still `null` when `getSubscribingMethods` is called, it should be initialized to this array, and a deprecation warning should be triggered
3. To avoid the deprecation warning and prepare for the future, `SonataCoreExtension` should set it to the configured value
4. In the configuration, the value should default to `array('json', 'xml', 'yml')`

ahmetakbn: @greg0ire do you mean that it will be configurable from a config file in the future?
So I need to add something like this:
```php
        $node
            ->children()
                ->arrayNode('serializer')
                    ->addDefaultsIfNotSet()
                    ->children()
                        ->arrayNode('formats')
                            ->prototype('scalar')->end()
                            ->defaultValue(array('json', 'xml', 'yml'))
                        ->end()
                    ->end()
                ->end()
            ->end()
        ;
```

Or just set it in the `SonataCoreExtension` like this:
```php
BaseSerializerHandler::setFormats($config['serializer']['formats']);
```
And since we are adding a deprecating warning why do we avoid from its message. Because if we are going to change it in the future shouldn't we show this warning message to tell them to change it, or configure it from configuration file?

greg0ire: > Or just

Both, I think.

> And since we are adding a deprecating warning why do we avoid from its message.

That's in case the class is used directly, without using the dependency injection at all (not sure this is possible though).

ahmetakbn: > Both, I think.

Oh, I copied pasted but forgot to change the code. After the 'Or Just' the code should be like this: 
```php
BaseSerializerHandler::setFormats(array('json', 'xml', 'yml'));
```
But anyway I got my answer.
> not sure this is possible though

As I know this is not possible. But I will add the deprecating message anyway.

ahmetakbn:
ping @greg0ire @Soullivaneuh @OskarStark

DependencyInjection/SonataCoreExtension.php
```diff
@@ -153,6 +156,14 @@ public function registerFlashTypes(ContainerBuilder $container, array $config)
         $container->setDefinition($identifier, $definition);
     }
 
+    /**
+     * @param array $config
+     */
+    public function configureSerializerFormats($config)
+    {
+        BaseSerializerHandler::setFormats($config['serializer']['formats']);
```
greg0ire: Perfect! :+1:

Resources/doc/reference/serialization.rst
```diff
@@ -7,7 +7,22 @@ Serialization
 Custom handlers
 ---------------
 
-The bundle comes with a ``BaseSerializerHandler`` to let you customize your serialized entities; this handler is used to serialize/deserialize an entity to/from its id, but you remain free to create your own handler for your specific needs.
+The bundle comes with a ``BaseSerializerHandler`` to let you customize your serialized entities;
+this handler is used to serialize/deserialize an entity to/from its id within the defaults
+formats ('json', 'xml', 'yml').
+
+The serializer default formats are configurable. You can change it from the configuration file.
```
greg0ire: it => them

Tests/Serializer/BaseSerializerHandlerTest.php
```diff
@@ -19,7 +19,7 @@
  */
 final class BaseSerializerHandlerTest extends \PHPUnit_Framework_TestCase
 {
-    public function testGetSubscribingMethods()
+    public function testGetSubscribingMethodsWithDefaultFormats()
```
greg0ire: Should be added to the `@legacy` group, right ? Don't you get deprecation notices when running that? Also, add a `NEXT_MAJOR` label to say this should call `setFormats`

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [3.x]
+### Added
+- Added static `setFormats` and `addFormat` methods to `BaseSerializerHandler` class
```
Soullivaneuh: Please do not add the changelog on the file.

DependencyInjection/Configuration.php
```diff
@@ -108,4 +109,26 @@ private function addFlashMessageSection(ArrayNodeDefinition $node)
...
+     * @param ArrayNodeDefinition $node
+     */
+    private function addSerializerFormats(ArrayNodeDefinition $node)
+    {
+        $node
+            ->children()
+                ->arrayNode('serializer')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('formats')
```
OskarStark: i would like to have an `->info()` node here with a short description

CHANGELOG.md
ahmetakbn: Sorry, I didnt get it. Where should I add the changelog information?

greg0ire: In the body of the PR

Tests/Serializer/BaseSerializerHandlerTest.php
ahmetakbn: Indeed, the last version should have this annotations. Check.

CHANGELOG.md
ahmetakbn: So, this change should be just deleted?

greg0ire: Yup. The goal is to avoid everyone editing the same line, creating conflicts again and again.

ahmetakbn: Got it. thanks ;)

ahmetakbn:
PR looks ready from my side, FYI.

Serializer/BaseSerializerHandler.php
```diff
@@ -35,15 +40,41 @@ public function __construct(ManagerInterface $manager)
...
      * {@inheritdoc}
      */
     public static function getSubscribingMethods()
     {
+        // NEXT_MAJOR : remove this block
+        if (null === static::$formats) {
+            static::$formats = array('json', 'xml', 'yml');
+            @trigger_error(
+                '$formats has been set to default array("json", "xml", "yml"). Setting $formats to a 
+                default array is deprecated since version 3.x and will be removed in 4.0. Use SonataCoreBundle
```
OskarStark: since `3.3`

/cc @Soullivaneuh ?

OskarStark:
Thank you @ahmetakbn after my comment this PR is `RTM` 👍

Serializer/BaseSerializerHandler.php
ahmetakbn: I changed to 3.0

OskarStark:
@greg0ire can you please give us a final review on this PR?

Thank you!

greg0ire:
Thanks @ahmetakbn !

fracsi:
If someone is not using `jms/serializer-bundle` then this PR breaks the application.

greg0ire:
Can you file an issue about that @fracsi ?

OskarStark:
@ahmetakbn can you support us here?

PR #243, opened 2016-04-29
Fix typo introduced in #238 (2.x branch)

fracsi:
Fixes: #241 in 2.x branch

PR #314, opened 2016-06-21
`choices_as_values` option typo

ossinkine:
The same as #243


Soullivaneuh:
The 2.x branch is not maintained anymore.

PR #316, opened 2016-06-25
Remove internal test classes from autoloader

core23:
I am targetting this branch, because this is a major BC break for all tests.

Related to https://github.com/sonata-project/dev-kit/issues/179

## Changelog

```markdown
### Removed
- Internal test classes are now excluded from the autoloader
```

## Subject

Tests for internal components shouldn't be loaded in production or extended by other bundles.

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -1,2 +1,6 @@
 UPGRADE FROM 3.X to 4.0
 =======================
+
+### Tests
+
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
Have you check if test classes are not used by other bundle for each PR?

For this one, I think we have at least one class used by others.

Soullivaneuh:
Should not be done on stable branch as the others?

PR #318, opened 2016-06-27
added test for AdapterCompilerPass

ahmetakbn:
### Subject

Improving tests for `SonataCoreBundle`

### DIFF
--- a/Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
+++ b/Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Tests\DependencyInjection\Compiler;
+
+use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractCompilerPassTestCase;
+use Sonata\CoreBundle\DependencyInjection\Compiler\AdapterCompilerPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @author Ahmet Akbana <ahmetakbana@gmail.com>
+ */
+final class AdapterCompilerPassTest extends AbstractCompilerPassTestCase
+{
+    public function registerCompilerPass(ContainerBuilder $container)
+    {
+        $container->addCompilerPass(new AdapterCompilerPass());
+    }
+
+    public function testDefinitionsAdded()
+    {
+        $coreModelAdapterChain = new Definition();
+        $this->setDefinition('sonata.core.model.adapter.chain', $coreModelAdapterChain);
+
+        $this->registerService('doctrine', 'foo');
+        $this->registerService('doctrine_phpcr', 'foo');
+
+        $this->compile();
+
+        $this->assertContainerBuilderHasServiceDefinitionWithMethodCall(
+            'sonata.core.model.adapter.chain',
+            'addAdapter',
+            array(new Reference('sonata.core.model.adapter.doctrine_orm'))
+        );
+
+        $this->assertContainerBuilderHasServiceDefinitionWithMethodCall(
+            'sonata.core.model.adapter.chain',
+            'addAdapter',
+            array(new Reference('sonata.core.model.adapter.doctrine_phpcr'))
+        );
+    }
+
+    public function testDefinitionsRemoved()
+    {
+        $coreModelAdapterChain = new Definition();
+        $this->setDefinition('sonata.core.model.adapter.chain', $coreModelAdapterChain);
+
+        $this->registerService('sonata.core.model.adapter.doctrine_orm', 'foo');
+        $this->registerService('sonata.core.model.adapter.doctrine_phpcr', 'foo');
+
+        $this->compile();
+
+        $this->assertContainerBuilderNotHasService('sonata.core.model.adapter.doctrine_orm');
+        $this->assertContainerBuilderNotHasService('sonata.core.model.adapter.doctrine_phpcr');
+    }
+}
--- a/DependencyInjection/Compiler/AdapterCompilerPass.php
+++ b/DependencyInjection/Compiler/AdapterCompilerPass.php
@@ -28,7 +28,11 @@ class AdapterCompilerPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        $definition = $container->getDefinition('sonata.core.model.adapter.chain');
+        if (!$container->has('sonata.core.model.adapter.chain')) {
+            return;
+        }
+
+        $definition = $container->findDefinition('sonata.core.model.adapter.chain');
 
         if ($container->has('doctrine')) {
             $definition->addMethodCall('addAdapter', array(new Reference('sonata.core.model.adapter.doctrine_orm')));


Soullivaneuh:
I would like to see [matthiasnoback/symfony-dependency-injection-test](https://github.com/matthiasnoback/SymfonyDependencyInjectionTest) implementation for this.

See also this WIP: https://github.com/sonata-project/SonataAdminBundle/pull/3932

greg0ire:
Related : http://elnur.pro/testing-repositories/

I don't think a **unit** test has much value in that case.

ahmetakbn:
> I don't think a unit test has much value in that case

@greg0ire Can you please elaborate?

greg0ire:
> @greg0ire Can you please elaborate?

Sure.

> First, you should unit test only the code that’s under your control. And you should mock only your own types, even thought sometimes it’s perfectly okay to mock third-party interfaces.

> Since repositories are usually chock full of third-party API calls, you can’t really control what code to write — you have to play by the rules of the APIs. Therefore unit testing them doesn’t make much sense.

In your case, the external API is the API of the container. Your test is very long because for every call to this API, you have to make a mocking equivalent.


> Second, repositories are at the boundaries of your beautiful application and the ugly outside world. A database is not on the side of your application; it’s outside of it. Repositories are the ones that connect the both worlds. Therefore mocking the APIs of the outside world or the ones connecting with it doesn’t make sense. On the other hand, since repositories’ interfaces belong to you, mocking repositories themselves makes sense.

Same goes for the container, it's not part of your application, it's part of the framework.

> Third, each database system has its own specific way of communicating with it. If you tie your unit tests to inner workings of a third-party component by basically recreating the functionality of that component in your mocks and feel really great about yourself and how smart you are, I’m happy for you. Just don’t whine that your tests are insanely hard to write, are brittle, and maintaining them is a nightmare.

Here, you are forced to mock every single call to the container, but a test should not depend on how the code is build : a tests should just ensure that given the right input, your code provides the right output.

> Fourth, just parsing and testing generated SQL should not give you any confidence that your code works because not everything can be derived from just SQL. You may be missing a required column in a table for SQL to work. A column or a table may have another name — after a database refactoring or for some other reason. Also databases have constraints that won’t get invoked unless you do a real database query.

Here, you have the same effect : you know you called the methods that should remove the services from the container, but you never check that they are actually removed.

> All that means that not only unit testing repositories is quite a complicated feat, those tests don’t even give you the confidence tests are written to give in the first place. How else would you call it if not masochism?

I have to agree. Try writing them with the integration test solution @Soullivaneuh recommended and you will see that you get smaller tests, and are more confident about the fact that indeed, it does work.

ahmetakbn:
Since we have the opportunity to check if our piece of code removes the service from the container, it makes sense to use that opportunity. Not about testing this class but the things you have mentioned and also are being criticised in the article you shared are not generally correct IMO.
Although there are some certain things in this article I am agree, there are also some parts I am totally disagree.
> First, you should unit test only the code that’s under your control. 

Agree

> And you should mock only your own types, even thought sometimes it’s perfectly okay to mock third-party interfaces.

Disagree. Because when I mock a third-party class that I use, I assume it works. It should have its own unit test. Actually by mocking this third-party code, my test will be testing only my code. 

> Since repositories are usually chock full of third-party API calls, you can’t really control what code to write — you have to play by the rules of the APIs. Therefore unit testing them doesn’t make much sense.

Of course I have to play by the rules of the APIs, because simply I am using it. But I am not unit testing them, I am mocking them to create my scenarios. 

> but a test should not depend on how the code is build

agree

> a tests should just ensure that given the right input, your code provides the right output

agree, but this is not the only purpose of test. Tests, for example should prevent you to not make changes that break your logic in future or miss any possible scenario. because test should also be responsible to correct all your scenarios.

> Here, you have the same effect : you know you called the methods that should remove the services from the container, but you never check that they are actually removed.

For this class we can assert that the service is removed from container. Thanks for dependency-test library. But we do not always have this option, therefore we mock the third-party and we assume that it works as expected. IMO, in test I should make it sure that I called the method with correct parameters. The response is the business of API. So I am not interested in what exactly it returns. Otherwise I would be testing the Api itself which does not make sense.

greg0ire:
That's right, we don't always have test cases we can extend for each external dependency, and can end up writing unit test that ensure that we call some methods we assume should work, and we can't check it actually works, because we are using mocks. And you're also right when pointing out that **unit** testing an external API makes no sense: it has already been **unit** tested. What you want to test mainly in doctrine repositories or symfony extensions is not the logic : there are very few if statements. What you actually want to test is that your software does the right API calls. So you won't be **unit** testing each call, but you will test that after n calls, you have the desired effect. Keep the unit test if there is some logic but make sure you havr an intégration test : you will see it is better suited to this kind of class.

greg0ire:
To go a bit further, I'd say that classes will little complexity do not deserve a unit test. For instance, the following method does not deserve one :

```php
function getName()
{
    return 'greg0ire';
}
```

If you have to change the test each time you change the method, there is a problem.

This one does not deserve one either:

```
function normalizeName($name)
{
    return strtoupper($name);
}
```

It would be unit testing php.

And finally, if there is no complexity in your class (no if statement and no loops, only one execution path), it probably does not deserve a unit test.

AdapterCompilerPass does have some if statements, which means there could be some value in unit tests, namely `testDoctrineAdapterIsEnabledWhenDoctrineIsPresent`, and 3 other tests. That would be more unitary, and could get abstracted into only 2 tests, with one data provider, named `testAdapterIsEnabledWhenServiceIsPresent` and `testAdapterDefinitionIsRemovedWhenServiceIsAbsent` . That would naturally lead to a refactoring of the pass itself, by introducing an array that would associate a service with the corresponding adapter, separating the logic from the API calls :

```php
protected $mapping = array(
    'doctrine' => 'sonata.core.model.adapter.doctrine_orm',
    'doctrine_phpcr' => 'sonata.core.model.adapter.doctrine_phpcr'
);

/**
 * For testing purposes
 */
public function setMapping(array $mapping)
{
    $this->mapping = $mapping;
}

public function process(ContainerBuilder $container)
{
    $definition = $container->getDefinition('sonata.core.model.adapter.chain');

    foreach ($this->mapping as $serviceId => $adapterServiceId) {
        if ($container->has($serviceId)) {
            $definition->addMethodCall('addAdapter', array(new Reference($adapterServiceId)));
        } else {
            $container->removeDefinition->($adapterServiceId);
        }
    }
}
```

So in short, if you're gonna write a unit test, make sure the unit test is easy
to write by refactoring the class if need be: no need for testing the same
logic you used for `doctrine` also works for `doctrine_phpcr`. If you're not
gonna do that, then write an integration test: you will still be able to test
the small piece of logic that is in your class, but will also be able to ensure
that the actual result is what is desired.

ahmetakbn:
@greg0ire thanks. I think we can also talk about this on concrete test codes. Would be really nice to be awared if sometime I miss something and over test some codes unnecessarily.

greg0ire:
By all means, let's talk about this on concrete cases : it will be easier for us to reason and understand how to choose the type of tests that are needed! Anyway, a big thank you for the time you spend writing these tests!

greg0ire:
In general, I have come to the realization that writing tests afterwards hardly leads to good tests : have a look at [the tests I wrote for the breadcrumbs builder](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Tests/Admin/BreadcrumbsBuilderTest.php). They're horrible, because I wrote them to ensure I did not break anything and thus wrote test without refactoring anything in the breadcrumbs builder. Then and only then I extracted the breadcrumbs builder into its own class, and knew it still worked, but… yikes, it's ugly! That's what I want to avoid, but sometimes you have no choice but to follow this process because there were zero tests : 

1. Write tests without refactoring the code.
2. Refactor the code without changing the tests, to make sure you don't break anything.
3. Refactor the tests without changing the code, to make sure you don't break anything.

Doing TDD, on the other hand, leads to better design from the start, and to clearer and easier to write tests.

ahmetakbn:
@Soullivaneuh changed test to use the library you mentioned.

Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $coreModelAdapterChain = new Definition();
+        $this->setDefinition('sonata.core.model.adapter.chain', $coreModelAdapterChain);
+
+        $this->compile();
+
+        $this->setExpectedException(
+            'Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException',
+            'You have requested a non-existent service "sonata.core.model.adapter.chain.doctrine_orm".'
+        );
+        $this->container->findDefinition('sonata.core.model.adapter.chain.doctrine_orm');
```
greg0ire: You can use `assertContainerBuilderNotHasService` directly ;)

Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $coreModelAdapterChain = new Definition();
+        $this->setDefinition('sonata.core.model.adapter.chain', $coreModelAdapterChain);
+
+        $this->compile();
+
+        $this->setExpectedException(
+            'Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException',
+            'You have requested a non-existent service "sonata.core.model.adapter.chain.doctrine_phpcr".'
+        );
+        $this->container->findDefinition('sonata.core.model.adapter.chain.doctrine_phpcr');
```
greg0ire: To be changed too

Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
greg0ire: This should fix the build, BTW :)

ahmetakbn:
integration test fails stupidly. Can you please take a look?

greg0ire:
> integration test fails stupidly. Can you please take a look?

See my comment ;)

ahmetakbn:
@greg0ire thanks

Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
```diff
@@ -0,0 +1,66 @@
...
+        $coreModelAdapterChain = new Definition();
+        $this->setDefinition('sonata.core.model.adapter.chain', $coreModelAdapterChain);
+
+        $this->registerService('sonata.core.model.adapter.doctrine_orm', 'foo');
+        $this->registerService('sonata.core.model.adapter.doctrine_phpcr', 'foo');
+
+        $this->compile();
+
+        $this->assertContainerBuilderNotHasService('sonata.core.model.adapter.doctrine_orm');
+        $this->assertContainerBuilderNotHasService('sonata.core.model.adapter.doctrine_phpcr');
```
greg0ire: Beautifully simple, thanks for rewriting this @ahmetakbn !

ahmetakbn: Yes, I also enjoyed while coding, was simple :)

greg0ire:
Thanks again @ahmetakbn !

Tests/DependencyInjection/Compiler/AdapterCompilerPassTest.php
greg0ire: That's what integration tests cases are all about, it's a shame this is not part of Symfony, and it's also a shame that they are rarely provided by libraries, because they avoid setting up a lot of boilerplate that can make the tests hard to read.

OskarStark:
Good job @ahmetakbn 🍻

PR #319, opened 2016-06-29
Remove internal test classes from autoloader

core23:
I am targetting this branch, because we some bundles have dependencies (e.g. https://github.com/sonata-project/SonataAdminBundle/pull/3977) to the tests.

Rebase of https://github.com/sonata-project/SonataCoreBundle/pull/316

Related to https://github.com/sonata-project/dev-kit/issues/179

## Changelog

```markdown
### Removed
- Internal test classes are now excluded from the autoloader
```

## Subject

Tests for internal components shouldn't be loaded in production or extended by other bundles.

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -1,2 +1,6 @@
 UPGRADE FROM 3.X to 4.0
 =======================
+
+### Tests
+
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
Should be on minor. If other bundle have dependencies, it should be resolved first.

core23:
If we do this on minor, projects could break because we remove stuff someone could use in their old version.

Soullivaneuh:
> we remove stuff someone could 

This someone guy should not extend test classes at all.

In this case, this should be the same for the other projects. But they was merged on minor branches.

PR #327, opened 2016-07-25
Choice type inversion and duplicate removes for Symfony 2.8+

cassianotartari:
I am targetting this branch, because `sonata_type_color_selector` wasn't compatible with Symfony 2.8+.  Need to invert array and remove duplicates to be according with Symfony 2.8+

Closes  https://github.com/sonata-project/SonataAdminBundle/issues/4016
## Changelog

``` markdown
### Added
- Added a Symfony 2.8+ test under `configureOptions` and filled choices with exchanged and duplicate removed colors.

### Fixed
- Symfony 2.8+ compatibility. As is the actual code the color saved will be the color name instead of its color html code. And selection option label is html code. To fix this I inverted the array and eliminated duplicates because aqua/cyan and magenta/fuchsia are the same color code.
```

- [x] Update the tests
## Subject

Exchanging all keys with their associated values in an array and removing duplicates to be compatible with Symfony 2.8+. Related to https://github.com/sonata-project/SonataAdminBundle/issues/4016


Form/Type/ColorSelectorType.php
```diff
@@ -33,22 +33,41 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
-                Colors::GREEN,
-                Colors::ORANGE,
-                Colors::PINK,
-                Colors::PURPLE,
-                Colors::RED,
-                Colors::WHITE,
-                Colors::YELLOW,
-            ),
-        ));
+        if(method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
```
greg0ire: Please add the missing space after `if` and `git commit --amend`, this should fix the failing build.

greg0ire:
The build fails for symfony > 2.7

greg0ire:
You have to change this test : https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Tests/Form/Type/ColorSelectorTypeTest.php#L115

Form/Type/ColorSelectorType.php
```diff
@@ -33,22 +33,41 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
+                    Colors::ORANGE,
+                    Colors::PINK,
+                    Colors::PURPLE,
+                    Colors::RED,
+                    Colors::WHITE,
+                    Colors::YELLOW,
+                ),
+            ));
+        } else { // SF <2.8 BC
+            $resolver->setDefaults(array(
```
core23: Please don't copy & paste stuff!

core23:
Ping @cassianotartari

cassianotartari:
sorry, totally forget this. I can't see it today, I'll schedule it for Monday morning.

cassianotartari:
How about now @core23 ?

Form/Type/ColorSelectorType.php
```diff
@@ -49,6 +49,13 @@ public function configureOptions(OptionsResolver $resolver)
                 Colors::YELLOW,
             ),
         ));
+
+        // SF >2.8
+        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+            $resolver->setDefaults(array_merge($resolver->getDefinedOptions(), array(
+                'choices' => array_unique(array_flip(Colors::getAll())),
+            )));
+        }
```
core23: We need the `choices_as_values` option too.

core23: Personal I would prefer something like:
```php
$defaults = array(
//...
);

if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) { 
   $defaults['choices'] = array_flip($defaults['choices']);
   $defaults['choices_as_values'] = true;
}

$resolver->setDefaults($defaults);

```

cassianotartari: ``choices_as_values``  is [deprecated](http://symfony.com/doc/current/reference/forms/types/choice.html#choices-as-values) in Symfony 3.x. Should I add this option? About the personal code it could be done, just clarify the ``choices_as_values`` option first.

core23: Is it deprecated, because the keys and values are flipped by default in symfony 3. For symfony 2.8+ we must declare it explicitly to stay BC.

OskarStark:
@cassianotartari travis is still failing, please have a look

cassianotartari:
@OskarStark I didn't write any tests yet. I'll try do it in a few minutes.

OskarStark:
@greg0ire why do we get the `getMock() is deprecated output?`

greg0ire:
b/c we use a recent version of phpunit on some build, and an old on others. Just disregard this particular output (more easily said than done, I know).

cassianotartari:
Done?

OskarStark:
@greg0ire can you please give this a final review?

Form/Type/ColorSelectorType.php
```diff
@@ -48,7 +48,15 @@ public function configureOptions(OptionsResolver $resolver)
                 Colors::WHITE,
                 Colors::YELLOW,
             ),
-        ));
+        );
+
+        // SF >2.8
```
greg0ire: There should be a NEXT_MAJOR indication here

Form/Type/ColorSelectorType.php
```diff
@@ -48,7 +48,15 @@ public function configureOptions(OptionsResolver $resolver)
...
+        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+            $defaults['choices'] = array_unique(array_flip(Colors::getAll()));
```
greg0ire: Should we edit the Colors class and flip all options and remove duplicates on NEXT_MAJOR ?

cassianotartari: Like in the test file?
```php
// NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
```

Tests/Form/Type/ColorSelectorTypeTest.php
```diff
@@ -268,6 +268,152 @@ public function testGetDefaultOptions()
...
+                'teal' => '#008080',
+                'thistle' => '#D8BFD8',
+                'tomato' => '#FF6347',
+                'turquoise' => '#40E0D0',
+                'violet' => '#EE82EE',
+                'wheat' => '#F5DEB3',
+                'white' => '#FFFFFF',
+                'whitesmoke' => '#F5F5F5',
+                'yellow' => '#FFFF00',
+                'yellowgreen' => '#9ACD32',
```
greg0ire: Don't test the data, test the logic (test the presence of one key is enough, with sth like arrayContainsKey)

Form/Type/ColorSelectorType.php
cassianotartari: Like in the test file?
```php
// NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
```

greg0ire: Yes

Form/Type/ColorSelectorType.php
cassianotartari: My knowledge doesn't understand the impacts of edit ``Colors`` to other branchs.

greg0ire: This would be a BC-break, and would therefore require a BC-break handling mechanism : 
- introducing an option to make `getAll()` return a flipped array of unique_key
- triggering a deprecation warning when the option is not set
- removing the option on `NEXT_MAJOR`

greg0ire: > My knowledge doesn't understand the impacts of edit Colors to other branchs.

Other packages you mean. I don't know either, but that's not a problem as long as we have a deprecation system : people will be warn, follow the upgrade procedure and be safe.

cassianotartari: I think I will say a big bullshit right now. Coding without really understand what should be done. It would be something like this?
```php
    /**
     * Return the list of colors.
     *
     * @return array
     */
    public static function getAll($isInverted = true)
    {
        if (func_num_args() == 0) {
            @trigger_error(
                'Omit $isInverted argument of '.__METHOD__.' is deprecated since 3.3, to be removed in 4.0.',
                E_USER_DEPRECATED
            );
        }
        if($isInverted) {
                return array(
                        'aliceblue' => self::ALICEBLUE,
                        'antiquewhite' => self::ANTIQUEWHITE,
                        //....
                );
        }
        return array(
            self::ALICEBLUE => 'aliceblue',
            self::ANTIQUEWHITE => 'antiquewhite',
            //....
        );
    }
```

Tests/Form/Type/ColorSelectorTypeTest.php
cassianotartari: I don't get. Shouldn't we test if array is inverted and without duplicates? I'm just trying to understand, I coded based on the initial test.

Form/Type/ColorSelectorType.php
greg0ire: No no, it's not bull :hankey: , it should indeed look like that a lot. Things to change from your code snippet : 

- use `array_flip()` to avoid repetition
- throw a deprecation also `$isInverted` is set to `false` (or `true` I don't remember which one is good), because if it is it means that people are not ready to upgrade (I was wrong when saying "when the option is not set")
- make the `NEXT_MAJOR` todo list as small as possible (removing a code block and be finished with the todo list for next major would be best).

Tests/Form/Type/ColorSelectorTypeTest.php
greg0ire: Shouldn't we test if array is inverted and without duplicates? Yes. Test exactly that. Here you are testing that the array is equal to some hardcoded value, when you could be quite sure it is inverted if there is a "yellowgreen" key and no "#9ACD32" key. Duplication can also be testing by comparing the length of the array with the length of `array_unique`'d array. When someone simply want to add a new color to the array, they shouldn't have to edit the tests IMO.

Form/Type/ColorSelectorType.php
cassianotartari: :smile: 
Maybe like this, if we think that color code should be the key of array (2.8+) invert the array should be done just for bellow Symfony versions.
```php
    /**
     * Return the list of colors.
     *
     * @return array
     */
    public static function getAll($isInverted = false)
    {
        $colors = array(
            self::ALICEBLUE => 'aliceblue',
            self::ANTIQUEWHITE => 'antiquewhite',
            //...
        );

        // NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
        if($isInverted) {
            @trigger_error(
                'The $isInverted argument of '.__METHOD__.' is deprecated since 3.3, to be removed in 4.0.',
                E_USER_DEPRECATED
            );
            return array_flip($colors);
        }
        
        return $colors;
    }
```
And we should remove the 2 duplicate ``const`` colors ``#FF00FF`` (MAGENTA/FUCHSIA) and ``#00FFFF`` (AQUA/CYAN)

greg0ire: > invert the array should be done just for bellow Symfony versions.

My thoughts exactly! That way you have a much smaller PR to do in NEXT_MAJOR

> And we should remove the 2 duplicate const colors #FF00FF (MAGENTA/FUCHSIA) and #00FFFF (AQUA/CYAN)

Hmm… wouldn't that be a BC-break? If yes, they should be added back to the flipped array, right ?

cassianotartari: > Hmm… wouldn't that be a BC-break? If yes, they should be added back to the flipped array, right ?

Good point... The colors are the same but with different names. We could remove than or make an:
```php
        // NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
        if($isInverted) {
            @trigger_error(
                'The $isInverted argument of '.__METHOD__.' is deprecated since 3.3, to be removed in 4.0.',
                E_USER_DEPRECATED
            );
            $colors[self::FUCHSIA] = 'fuchsia';
            $colors[self::CYAN] = 'cyan';
            return array_flip($colors);
        }
```

greg0ire: that would be perfect!

cassianotartari: I wrote inverted the "right" way in the code snippet before, so it would be like:
```php
    /**
     * Return the list of colors.
     *
     * @return array
     */
    public static function getAll($isInverted = false)
    {
        $colors = array(
            'aliceblue' => self::ALICEBLUE,
            'antiquewhite' => self::ANTIQUEWHITE,
            //...
        );

        // NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
        if($isInverted) {
            @trigger_error(
                'The $isInverted argument of '.__METHOD__.' is deprecated since 3.3, to be removed in 4.0.',
                E_USER_DEPRECATED
            );
            $colors[self::FUCHSIA] = 'fuchsia';
            $colors[self::CYAN] = 'cyan';
            return array_flip($colors);
        }
        
        return $colors;
    }
```
After I'll take a look to the tests.

greg0ire: The old way should be the default. So the default value should be `true`, right?

cassianotartari: Yeah, right. I was hangry, coming back from lunch, now I will finish this.

Tests/Form/Type/ColorSelectorTypeTest.php
cassianotartari: Would be something like this:
```php
    public function testGetDefaultOptions()
    {
        $type = new ColorSelectorType();

        $this->assertSame('sonata_type_color_selector', $type->getName());
        $this->assertSame(
            method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix') ?
                'Symfony\Component\Form\Extension\Core\Type\ChoiceType' :
                'choice',
            $type->getParent()
        );

        FormHelper::configureOptions($type, $resolver = new OptionsResolver());

        $options = $resolver->resolve();
        
        // NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
            $this->assertTrue($options['choices_as_values']);
            $this->assertArrayHasKey('yellowgreen', $options['choices']);
            $this->assertCount(count($options['choices']), array_unique($options['choices']));
        } else {
            $this->assertArrayHasKey(Colors::YELLOWGREEN, $options);
        }
    }
```

cassianotartari: How do I work with:
```bash
Remaining deprecation notices (2)

The $isInverted argument of Sonata\CoreBundle\Color\Colors::getAll is deprecated since 3.3, to be removed in 4.0: 2x
    1x in ColorSelectorTypeTest::testGetDefaultOptions from Sonata\CoreBundle\Tests\Form\Type
    1x in ColorSelectorTypeTest::testBuildForm from Sonata\CoreBundle\Tests\Form\Type
```
?

greg0ire: You must add the `@group legacy` annotation to the test method doc block to let the phpunit deprecation handler understand that this warning is internal and you are testing the legacy way of doing things, so this is OK. Use separate methods for testing the new way and the legacy way.

cassianotartari: > Use separate methods for testing the new way and the legacy way.

I don't get it.  One ``testGetDefaultOptions()`` without the ``if 2.8`` block and other ``testGetDefaultIsInvertedOptions()`` with ``if`` block ?

greg0ire: Yes. That way you can mark one with the annotation, and not the other, which shouldn't. `testGetDefaultOptionsLegacy` might be a better name, BTW.

cassianotartari: But ``Colors::getAll()`` is always called so both methods will trigger error. Even ``testBuildForm()`` will need ``@group legacy`` annotation.

This appears to not be right:
```php
    /**
     * @group legacy
     */
    public function testGetDefaultOptions()
    {
        $type = new ColorSelectorType();

        $this->assertSame('sonata_type_color_selector', $type->getName());
        $this->assertSame(
            method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix') ?
                'Symfony\Component\Form\Extension\Core\Type\ChoiceType' :
                'choice',
            $type->getParent()
        );

        FormHelper::configureOptions($type, $resolver = new OptionsResolver());

        $options = $resolver->resolve();
        
        // NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
        if (!method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
            $this->assertArrayHasKey(Colors::YELLOWGREEN, $options);
        }
    }
    
    /**
     * @group legacy
     */
    public function testGetDefaultOptionsLegacy()
    {
        $type = new ColorSelectorType();

        $this->assertSame('sonata_type_color_selector', $type->getName());
        $this->assertSame(
            method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix') ?
                'Symfony\Component\Form\Extension\Core\Type\ChoiceType' :
                'choice',
            $type->getParent()
        );

        FormHelper::configureOptions($type, $resolver = new OptionsResolver());

        $options = $resolver->resolve();
        
        // NEXT_MAJOR: Remove this "if" (when requirement of Symfony is >= 2.8)
        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
            $this->assertArrayHasKey('choices_as_values', $options);
            $this->assertTrue($options['choices_as_values']);
            $this->assertArrayHasKey('yellowgreen', $options['choices']);
            $this->assertCount(count($options['choices']), array_unique($options['choices']));
        }
    }
```

Form/Type/ColorSelectorType.php
```diff
@@ -33,7 +33,7 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefaults(array(
+        $defaults = array(
             'choices' => Colors::getAll(),
```
greg0ire: There needs to be a new argument there. This argument should come from the a configurable container parameter.
In the unit tests, this argument should be set to false so that you don't have to mark all tests with the legacy annotation. The user should be able to do the following and not get any deprecation notice : 

```
sonata_core:
   my_parameter: false # defaults to true since defaulting to false would not be BC
```

cassianotartari: Now I got totally lost :sob:

greg0ire: Let me re-explain then.

# Use case 1 : the user just uses the bundle

1. They upgrade to the next minor version, that contains your PR
2. They see a deprecation notice, that comes from the SonataCoreExtension class
3. They ignore it, it works just like before
4. They prepare for the next upgrade
5. They set the parameter to `false`, the new codepath is used, and the deprecation notice disappears
6. They change their code to accomodate the fact that there are two options less now.
7. They do the major upgrade happily


# Use case 2 : the user directly calls Colors::getAll() from their code

1. They upgrade to the next minor version, that contains your PR
2. They see a deprecation notice, that comes from the Colors class
3. They ignore it, it works just like before
4. They prepare for the next upgrade
5. They call the method like this : Colors::getAll(false)
6. They change their code to accomodate the fact that there are two options less now.
7. They do the major upgrade happily

Right now you have just covered "Use case 2", which is kind of secondary. You still need to cover Use case 1, which is the most important.

Is any step unclear?

cassianotartari: I really did my best to understand here, read it 4 times, smells bad here :smiley:  ...  Keep not understanding...

Independent of Symfony version the ``configureOptions`` will trigger the deprecated message because ``Colors::getAll()`` (``if ($isInverted)`` ``true``) is always called. I did not catch how to prevent this using Symfony 2.8+

greg0ire: > I really did my best to understand here, read it 4 times, smells bad here :smiley: ... Keep not understanding...

What step in particular? I think you should think hard about the goals : 

- having the user be able to upgrade to next **minor** and have everything work
- having the user be able to adopt "the new way" through configuration and get rid of deprecation messages
- having the tests run "the new way" in 90% of the cases, and use the 10% to test that yeah, you do get a deprecation notice when appropriate

> Independent of Symfony version the configureOptions will trigger the deprecated message because Colors::getAll() (if ($isInverted) true) is always called. I did not catch how to prevent this using Symfony 2.8+

I think you should change `Colors::getAll()` to `Colors::getAll($configurableParameterContainer)` in the code. In the test, set this parameter to false and you will be able to prevent this, right?

Color/Colors.php
```diff
@@ -167,150 +167,162 @@ class Colors
      *
      * @return array
      */
-    public static function getAll()
+    public static function getAll($isInverted = true)
```
core23: This is a BC break!

greg0ire: Is it?

greg0ire: You're right, [it is](https://3v4l.org/eRFvW) :scream:

greg0ire: What's the solution? Using `func_get_args()` ?

core23: This could be a solution, but it is pretty ugly...

core23: IMHO flipping the elements inside the method is also a BC break. `configureOptions` should be the right place to flip the elements, because we pass the options to symfony who can handle the flipped elements correctly.

greg0ire: > IMHO flipping the elements inside the method is also a BC break.

That's why I was suggesting doing 2 deprecation notices (see the 2 use cases). If we flip something, we must keep in mind that we will remove code in next major, and that when this code is removed, it would be great to have no flip at all left, don't you think?

core23:
Ping @cassianotartari

Color/Colors.php
cassianotartari: @greg0ire and @core23 what should I do?

Move ``array_flip`` to ``ColorSelectorType->configureOptions()`` ?

greg0ire: I think the goal in the end, in the next major is to have **no** `array_flip` at all. This means that there should be flipped colors + a call to `array_flip` in the deprecated path of `getAll()`, exactly like you did. @core23 , please comment on that.

core23: We can't do this in this method, because it is a BC break. So we must handle it in the calling method in `ColorSelectorType`.

greg0ire: @cassianotartari manually flipped the colors, and then the default, deprecated path flips them again, which means if you upgrade… you will get the same as before, so no BC-break, right?

core23: The method signature itself is a BC break.

greg0ire: It is, but for extension. Can we consider this BC-break ok? I'm not even sure people are using this class directly, so I doubt they extend it. If you think otherwise, then the only solution I see is creating a new method and deprecating this one.

core23: What about deprecate this class and move the colors to the `ScolorSelectorType`?

greg0ire: Might be a good idea, `Colors` does not seem to be used on its own. Or maybe just move the data (because hey, this class is basically just data) to a YML file (or any other suitable format), and load that file in the type?

cassianotartari:
I did a mess right now and I don't know how to fix it... Added the ``pt_BR`` translation inside this PR. How can I remove it?

OskarStark:
> I did a mess right now and I don't know how to fix it... Added the pt_BR translation inside this PR. How can I remove it?

`git rebase -i HEAD~2`, afterwards change `pick` to `drop` or `d` in front of the commit you want to remove!

greg0ire:
Nevermind my comment

cassianotartari:
> git rebase -i HEAD~2, afterwards change pick to drop or d in front of the commit you want to remove and git push --force

```bash
git push --force
```

Throws:
```bash
$ git push --force
warning: push.default is unset; its implicit value has changed in
Git 2.0 from 'matching' to 'simple'. To squelch this message
and maintain the traditional behavior, use:

  git config --global push.default matching

To squelch this message and adopt the new behavior now, use:

  git config --global push.default simple

When push.default is set to 'matching', git will push local branches
to the remote branches that already exist with the same name.

Since Git 2.0, Git defaults to the more conservative 'simple'
behavior, which only pushes the current branch to the corresponding
remote branch that 'git pull' uses to update the current branch.

See 'git help config' and search for 'push.default' for further information.
(the 'simple' mode was introduced in Git 1.7.11. Use the similar mode
'current' instead of 'simple' if you sometimes use older versions of Git)

fatal: You are not currently on a branch.
To push the history leading to the current (detached HEAD)
state now, use

    git push origin HEAD:<name-of-remote-branch>
```

boezerooij:
Any news? I would like to use the sonata_type_color_selector

OskarStark:
@cassianotartari can you please fix travis? afterwards we can merge this

core23:
> afterwards we can merge this

Nope, there is still a BC break in it!

Color/Colors.php
greg0ire: @verheyenkoen @core23 I have an even better (I hope) idea. How about implementing a type that creates `<input type="color" />`, and deprecate this one? Not sure why sf didn't do it itself already.

greg0ire: https://github.com/symfony/symfony/issues/22596

greg0ire:
Merged #411 instead, which is probably what you did originally @cassianotartari , sorry for that. We still need someone to provide a new color type in case you or @verheyenkoen want to do it. That way we could drop the `Colors` type completely, and avoid this inversion issue.

verheyenkoen:
@greg0ire You mean we shouldn't wait for that type in Symfony?

greg0ire:
We probably shouldn't, since people using old versions of sf wouldn't be able to use it, and we still support sf 2.8. If sf does provide the type someday, then we will just conditionally deprecate ours.

PR #328, opened 2016-07-29
Deprecate exporter

### LINKED ISSUE #50: [Feature request] CSV export options
Situation
=======

Currently the ``Sonata\CoreBundle\Exporter\Exporter`` does not have any options for the column delimiter or the enclosing character if ``$format = csv`` is used.
Instead its using the following hardcoded:

``$writer      = new CsvWriter('php://output', ',', '"', "", true, true);``

Use Case
========
Theoretically the file is a correct CSV-File using comma as seperator but
opening a CSV-File generated by this exporter under Windows with Mircosoft Excel can cause that Excel does not manage to separate the columns correctly (depends on the OS system configurations and/or the Excel version and its configuration)

Prospect
=======
The exporter should have another param for the ``getResponse`` method which could be called ``$formatOptions``. That could be an array of options for the selected export format. That would imply that the SonataAdminBundle needs to be adjusted as well and provide the configured values for the exported format.

It seems that CSV is the only export format that needs  to be configureable at the moment. This makes it quite difficult for me to suggest where the best place would be to set up the options for the CSV export.

Admin class ?
config.yml ?
both ?

greg0ire:
I am targetting this branch, because this is BC

Closes #50

## Changelog

```markdown
### Deprecated
- Exporter class and service : use equivalents from `sonata-project/exporter` instead.
```

## To do

- [x] Wait for 1.6.0 to be released

## Subject

This deprecates the export class and service

### DIFF
--- a/Exporter/Exporter.php
+++ b/Exporter/Exporter.php
@@ -18,6 +18,15 @@ use Exporter\Writer\XlsWriter;
 use Exporter\Writer\XmlWriter;
 use Symfony\Component\HttpFoundation\StreamedResponse;
 
+@trigger_error(
+    'The '.__NAMESPACE__.'\Exporter class is deprecated since version 3.x and will be removed in 4.0.'.
+    ' Use Exporter\Exporter instead',
+    E_USER_DEPRECATED
+);
+
+/**
+ * NEXT_MAJOR: remove this class, and the dev dependency.
+ */
 class Exporter
 {
     /**
--- a/Resources/config/exporter.xml
+++ b/Resources/config/exporter.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- NEXT_MAJOR : remove this file -->
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="sonata.core.exporter" class="Sonata\CoreBundle\Exporter\Exporter"/>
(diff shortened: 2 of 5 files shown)


greg0ire:
@sonata-project/contributors minor? patch?

greg0ire:
@sonata-project/contributors please review this

OskarStark:
LGTM 👍

OskarStark:
btw. should we increase the dependency to 1.6.0 `

greg0ire:
> btw. should we increase the dependency to 1.6.0

I don't think so, it would be a BC-break, right?

core23:
Should be minor, because we deprecate stuff: http://semver.org/#spec-item-7

Exporter/Exporter.php
```diff
@@ -18,6 +18,15 @@
...
+@trigger_error(
+    'The '.__NAMESPACE__.'\Exporter class is deprecated since version 3.x and will be removed in 4.0.'.
+    ' Use Exporter\Exporter instead',
+    E_USER_DEPRECATED
+);
+
+/**
+ * NEXT_MAJOR: remove this class, and the dev dependency.
+ */
 class Exporter
```
core23: Can we require the new lib and extend it here?

greg0ire: I don't think so. Wouldn't it be a BC break? see https://github.com/sonata-project/SonataCoreBundle/pull/328#issuecomment-237576469

core23:
>> btw. should we increase the dependency to 1.6.0

> I don't think so, it would be a BC-break, right?

Don't thinkt, that this is a break, because we require a higher minor version. People who want's to use the exporter lib <1.6, won't get the latest version of the core bundle. This is how composer works 👍 

Btw. we did this already on several bundles, when requiring the 3.1 version of the admin bundle:
https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json#L49

greg0ire:
Ok I'll do as you suggest.

greg0ire:
@core23 : I reverted to the previous version : `Exporter\Exporter` is `final`.

core23:
Okay :(

Then LGTM

PR #329, opened 2016-07-29
Silence internal deprecation notice

greg0ire:
I am targetting this branch, because it is BC

## Subject

Deprecation notice for internal deprecation should be considered normal, and be tested.


greg0ire:
ping @sonata-project/contributors , this is a no-brainer

Soullivaneuh:
This is not patch but pedantic for this case.

Thank you.

PR #336, opened 2016-08-19
Added option to override template location in AbstractWidgetTestCase

core23:
I am targetting this branch, because…

## Changelog

```markdown
### Added
- Added `AbstractWidgetTestCase::getTemplatePaths` method
```

## Subject

Should be the final PR for using the test case outside the bundle.

### DIFF
--- a/Test/AbstractWidgetTestCase.php
+++ b/Test/AbstractWidgetTestCase.php
@@ -60,6 +60,24 @@ abstract class AbstractWidgetTestCase extends TypeTestCase
      */
     protected function getEnvironment()
     {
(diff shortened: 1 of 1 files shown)


OskarStark:
I think we should add a test for the `AbstractWidgetTestCase` and ensure that `getTemplatePaths()` is working properly

what do you think guys? @greg0ire @core23

greg0ire:
Testception, but why not?

core23:
Don't know if we can test this, because we have a hardcoded class in the TestCase.

OskarStark:
> Don't know if we can test this, because we have a hardcoded class in the TestCase.

OK, as this change is not that complex we keep it as it is, thank you @core23

PR #356, opened 2016-11-05
Remove deprecated things

greg0ire:
I am targetting this branch, because this is full of (planned) BC-breaks

## Changelog

Deprecated things are already in the CHANGELOG

PR #363, opened 2016-12-04
Merge 3.x

core23:
Manual merge of 3.x into the master branch.


greg0ire:
There are conflicts. If you want to do this, you have to create a new branch from master, merge 3.x in that, resolve the conflicts, and make a PR from that.

core23:
Fixed the conflicts 👍 

> If you want to do this, you have to create a new branch from master, merge 3.x in that, resolve the conflicts, and make a PR from that.

That's what I've done.

core23:
Your PR https://github.com/sonata-project/SonataCoreBundle/pull/356 has made some problems. Next time we should merge the stable branch into the master before removing deprecated stuff.

greg0ire:
The problem is that there will always be new things on master, so I think the right thing to do is to merge regularly, so as not to have a gazillion conflicts to change each time. That or wait the last moment before removing deprecated stuff…

core23:
> The problem is that there will always be new things on master

That's right, but there were 25 commits which is very annoying. We should find a way to notify us via slack, if the `dev-kit` can't merge it.

greg0ire:
@Soullivaneuh please review this. Since you're the one who does these merges usually…

PR #364, opened 2016-12-06
Stop using private property

greg0ire:
I am targetting this branch, because this is BC (only tests are changing).

## Changelog

```markdown
### Fixed
- the abstract test case for testing widgets now works with Symfony 3.2
```

## To do
- [x] Find a clean way to detect symfony/twig-bridge 3.2

## Subject

This PR is necessary for https://github.com/sonata-project/SonataAdminBundle/pull/4216 to pass.

### DIFF
--- a/Test/AbstractWidgetTestCase.php
+++ b/Test/AbstractWidgetTestCase.php
@@ -35,6 +35,11 @@ abstract class AbstractWidgetTestCase extends TypeTestCase
     private $extension;
 
     /**
+     * @var TwigRenderer
+     */
+    private $renderer;
+
+    /**
      * {@inheritdoc}
      */
     protected function setUp()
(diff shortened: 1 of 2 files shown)


greg0ire:
The title of this PR feels a bit communist…

greg0ire:
@wouterj @Koc @fbourigault @cedricziel please review (and by that I mean approve :P)

composer.json
```diff
@@ -22,6 +22,7 @@
         "symfony/form": "^2.3.5 || ^3.0",
         "symfony/http-foundation": "^2.3 || ^3.0",
         "symfony/property-access": "^2.3 || ^3.0",
+        "symfony/security": "^2.3 || ^3.0",
```
Soullivaneuh: What is the relation between this and the twig property ?

greg0ire: It has to do with this line : https://github.com/sonata-project/SonataCoreBundle/pull/364/files#diff-ab34013c1c5e23c243a5c45874748062R52

greg0ire: This particular line can't work if the security component is not installed.

greg0ire: @Soullivaneuh : I moved this to a separate commit, to make it more clear.

jordisala1991: @greg0ire If it is only used by tests, shouldn't it be on require-dev? Same happens for the twig-bridge

greg0ire: It's a test that can be used outside the bundle (as explained in my commit message :P). BTW, this is why I said it's necessary to merge this if we want https://github.com/sonata-project/SonataAdminBundle/pull/4216 to be green one day

jordisala1991: well, can be used outside the project, but for testing purposes... still non-dev?

greg0ire: To sum up, if something is in the `Test` directory (as opposed to the `Tests` directory), its dependencies should be in `require`, because it is exposed to client projects.

Test/AbstractWidgetTestCase.php
```diff
@@ -49,10 +54,25 @@ protected function setUp()
             'Symfony\Component\Form\Extension\Csrf\CsrfProvider\CsrfProviderInterface',
         ), 'interface_exists');
 
-        $renderer = new TwigRenderer($this->getRenderingEngine(), $this->getMock(current($csrfProviderClasses)));
+        $this->renderer = new TwigRenderer($this->getRenderingEngine(), $this->getMock(current($csrfProviderClasses)));
+
+        $this->extension = new FormExtension($this->renderer);
+        $environment = $this->getEnvironment();
+
+        // TODO: remove the condition when dropping symfony/twig-bundle < 3.2
```
greg0ire: @sonata-project/contributors : can't use `NEXT_MAJOR` here, what should I do?

greg0ire:
@sonata-project/contributors , can you please merge this? It's critical because it blocks sonata-project/SonataAdminBundle#4216 , which in turn will block any new PR on SonataAdminBundle.

core23:
Thanks @greg0ire

greg0ire:
Thank you @core23 !

PR #376, opened 2017-01-13
removed useless class comments and fixed author PHPDoc

OskarStark:
I am targetting this branch, because this is BC.


Command/SonataDumpDoctrineMetaCommand.php
```diff
@@ -21,8 +21,6 @@
 
 /**
  * Return useful data on the database schema.
- *
- * Class SonataDoctrineUtils
```
greg0ire: Lol that wasn't even true

OskarStark:
Fixed CS

PR #384, opened 2017-02-03
Removed deprecated getMock in unit tests

core23:
I am targetting this branch, because this is for tests only.

## Changelog

```markdown
### Changed
 - `AbstractWidgetTestCase` uses `createMock` method
```

## Subject

The method `getMock` is deprecated.

### DIFF
--- a/Test/AbstractWidgetTestCase.php
+++ b/Test/AbstractWidgetTestCase.php
@@ -58,7 +58,7 @@ abstract class AbstractWidgetTestCase extends TypeTestCase
             'Symfony\Component\Form\Extension\Csrf\CsrfProvider\CsrfProviderInterface',
         ), 'interface_exists');
 
-        $this->renderer = new TwigRenderer($this->getRenderingEngine(), $this->getMock(current($csrfProviderClasses)));
+        $this->renderer = new TwigRenderer($this->getRenderingEngine(), $this->createMock(current($csrfProviderClasses)));
 
         $this->extension = new FormExtension($this->renderer);
         $environment = $this->getEnvironment();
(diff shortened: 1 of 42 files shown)


Test/AbstractWidgetTestCase.php
```diff
@@ -172,4 +172,20 @@ protected function getRenderingEngine()
...
     }
+
+    /**
+     * NEXT_MAJOR: Remove this method when dropping support for < PHPUnit 5.4.
+     *
+     * @param string $class
+     *
+     * @return \PHPUnit_Framework_MockObject_MockObject
+     */
+    protected function createMock($class)
```
greg0ire: Please make it final.

Tests/Form/Type/TypeTestCase.php
```diff
@@ -0,0 +1,34 @@
...
+ * @internal
+ */
+class TypeTestCase extends \Symfony\Component\Form\Test\TypeTestCase
+{
+    /**
+     * @param string $class
+     *
+     * @return \PHPUnit_Framework_MockObject_MockObject
+     */
+    protected function createMock($class)
```
greg0ire: You can make it final here too, but it's less of an issue because this class is not part of the public API.

Tests/PHPUnit_Framework_TestCase.php
```diff
@@ -0,0 +1,34 @@
...
+ * @internal
+ */
+class PHPUnit_Framework_TestCase extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @param string $class
+     *
+     * @return \PHPUnit_Framework_MockObject_MockObject
+     */
+    protected function createMock($class)
```
greg0ire: Same here, as you wish.

Test/AbstractWidgetTestCase.php
core23: 👎 because the original `createMock` method isn't final too: https://github.com/sebastianbergmann/phpunit/blob/5.4/src/Framework/TestCase.php#L1526

This would irritate someone.

greg0ire: You're right, nevermind that change.

Tests/Form/Type/TypeTestCase.php
core23: As you said, this `is not part of the public API`. So no reason to close this API.

OskarStark:
Thank you @core23 👍

PR #391, opened 2017-03-23
Improve error handling

greg0ire:

## Changelog

```markdown
### Added
- Options for `ImmutableArrayType` are now validated, which should yield better results.
```

## Subject

Options for `ImmutableArrayType` are quite peculiar, and misusing this type can
lead to somewhat cryptic error messages, like for instance "Undefined offset:
2", from `app/cache/classes.php`, which does not help at all.

### DIFF
--- a/Form/Type/ImmutableArrayType.php
+++ b/Form/Type/ImmutableArrayType.php
@@ -64,6 +64,21 @@ class ImmutableArrayType extends AbstractType
         $resolver->setDefaults(array(
             'keys' => array(),
         ));
+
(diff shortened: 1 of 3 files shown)


jordisala1991:
Seems like setAllowedValues has changed its API from SF2.3?

greg0ire:
> Seems like setAllowedValues has changed its API from SF2.3?

Indeed… not sure since when closures are supported either, I'll have to rework this.

jordisala1991:
Thanks @greg0ire

PR #396, opened 2017-04-07
WIP: Symfony33 compat form extension

yann-eugone:
Do not merge.

This PR is a test, trying to resolve https://github.com/sonata-project/SonataAdminBundle/issues/4447.

PR #399, opened 2017-04-20
Rely *explicitly* on FrameworkBundle's FormPass

nicolas-grekas:
Working alternative to #396
Fixing https://github.com/sonata-project/SonataAdminBundle/issues/4447
Needs https://github.com/symfony/symfony/pull/22481

### DIFF
--- a/DependencyInjection/Compiler/FormFactoryCompilerPass.php
+++ b/DependencyInjection/Compiler/FormFactoryCompilerPass.php
@@ -11,10 +11,10 @@
 
 namespace Sonata\CoreBundle\DependencyInjection\Compiler;
 
-use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-class FormFactoryCompilerPass implements CompilerPassInterface
+class FormFactoryCompilerPass extends FormPass
 {
     /**
      * {@inheritdoc}
@@ -41,6 +41,7 @@ class FormFactoryCompilerPass implements CompilerPassInterface
 
         // get factories
         $original = $container->getDefinition('form.extension');
+        parent::process($container);
 
         $factory = $container->getDefinition('sonata.core.form.extension.dependency');
         $factory->replaceArgument(1, $original->getArgument(1));
--- a/Tests/DependencyInjection/Compiler/FormFactoryCompilerPassTest.php
+++ b/Tests/DependencyInjection/Compiler/FormFactoryCompilerPassTest.php
@@ -52,27 +52,22 @@ final class FormFactoryCompilerPassTest extends AbstractCompilerPassTestCase
 
     public function testProcessWithContainerHasFormExtensionDefinition()
     {
-        $formExtension = new Definition();
-        $formExtension->addArgument('foo');
-        $formExtension->addArgument('bar');
-        $formExtension->addArgument('baz');
-        $formExtension->addArgument('caz');
+        $formExtension = new Definition(null, array(null, null, null, null));
         $this->setDefinition('form.extension', $formExtension);
 
-        $sonataFormExtension = new Definition();
-        $sonataFormExtension->addArgument('extension-foo');
-        $sonataFormExtension->addArgument('extension-bar');
-        $sonataFormExtension->addArgument('extension-baz');
-        $sonataFormExtension->addArgument('extension-caz');
+        $formType = new Definition();
+        $formType->addTag('form.type');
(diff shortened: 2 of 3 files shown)


DependencyInjection/Compiler/FormFactoryCompilerPass.php
```diff
@@ -11,10 +11,10 @@
 
 namespace Sonata\CoreBundle\DependencyInjection\Compiler;
 
-use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass;
```
greg0ire: The composer.json should probably be modified too, right?

nicolas-grekas: why?

greg0ire: Because there is no explicit requirement for the framework bundle in it I think

nicolas-grekas: There is no bundle with no requirement on fwb... That's why there argument can be "replaced" (they come from fwb already)

greg0ire: > There is no bundle with no requirement on fwb

I don't get what you're saying, but if we reference fwb somewhere in our code, then we should express what version constraints we have, right?

nicolas-grekas: I mean that all bundles have an implicit requirement on FrameworkBundle.
Take this line: `$container->getDefinition('form.extension')`
Where does this `form.extension` come from ? FrameworkBundle.
Thus, this PR adds no additional requirements at all.

greg0ire: I'm not saying this PR adds a requirement that was not here before, but since we are making things explicit, we should probably add the requirement for the sake of completeless, even more because IMO, if you reference a FQCN, you should require the package that provides it (apart maybe from testing packages). Right now it works because :

```
composer why symfony/framework-bundle
friendsofsymfony/rest-bundle   2.1.1    requires  symfony/framework-bundle (^2.7|^3.0)  
jms/serializer-bundle          1.2.1    requires  symfony/framework-bundle (~2.3|~3.0)  
nelmio/api-doc-bundle          2.13.1   requires  symfony/framework-bundle (~2.3|~3.0)  
sensio/framework-extra-bundle  v3.0.25  requires  symfony/framework-bundle (~2.3|~3.0)
```

nicolas-grekas: added

nicolas-grekas:
PR should be green.

Note that this is just a temporary patch: by replacing the `form.extension` service implementation, Sonata nullifies the benefit of lazy iteration as introduced in 3.3.

greg0ire:
> Note that this is just a temporary patch: by replacing the form.extension service implementation, Sonata nullifies the benefit of lazy iteration as introduced in 3.3.

Thanks a lot! Do you have a recommendation regarding the way to go? Maybe go on with #396 ?

nicolas-grekas:
Getting rid of this custom form extension... Or making it deal with iterator (without using iterator_to_array)

jordisala1991:
symfony/symfony#22481 is already merged, should we merge this now or wait to 3.3 release?

nicolas-grekas:
you can merge, there's no blocker for me

greg0ire:
Thanks @nicolas-grekas !

OskarStark:
@greg0ire can you please add a changelog to the PR header?

greg0ire:
@OskarStark done

DependencyInjection/Compiler/FormFactoryCompilerPass.php
```diff
@@ -41,6 +41,7 @@ public function process(ContainerBuilder $container)
 
         // get factories
         $original = $container->getDefinition('form.extension');
+        parent::process($container);
```
dbu: this is replacing arguments of the 'form.extension' service (aka $original). is that intended? in symfony 3.3 i get type mismatches because object (iterator?) instead of an array.

OskarStark: any comments on this @nicolas-grekas ?

nicolas-grekas: @dbu is this still an issue?

dbu: unfortunately i have no idea anymore how i produced this. in the cmf sandbox we don't see errors atm, so i guess we can leave it, if it makes sense semantically.

PR #402, opened 2017-05-14
Fix translation has to be not null for twig 2.0

xaben:
I am targeting this branch, because this fixes a bug with twig 2.0.

## Changelog

```markdown

### Fixed
- fixed Twig errors when using `sonata_template_box` without a translation domain

```

## Subject

Using `sonata_template_box` without a translation domain, results in passing a null value of `translationBundle` to Twig_Node constructor, which is no longer supported in Twig 2.0.

This PR resolves the issue by removing `translationBundle` if it is `null` from the passed parameters.

This is another solution for https://github.com/sonata-project/SonataMediaBundle/pull/1255

### DIFF
--- a/Tests/Twig/TokenParser/TemplateBoxTokenParserTest.php
+++ b/Tests/Twig/TokenParser/TemplateBoxTokenParserTest.php
@@ -21,7 +21,7 @@ class TemplateBoxTokenParserTest extends PHPUnit_Framework_TestCase
      * @dataProvider getTestsForRender
      *
      * @param bool            $enabled
-     * @param string          $source
+     * @param \Twig_Source    $source
      * @param TemplateBoxNode $expected
      *
      * @throws \Twig_Error_Syntax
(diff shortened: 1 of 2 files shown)


OskarStark:
Shouldn't this be handled in the code itself?

xaben:
@OskarStark Sorry I don't understand your comment. Just to be clear, due to https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Node.php#L41 all nodes passed to the constructor inside the array need to be a `\Twig_Node`. That's why null is no longer a valid option for `translationBundle`

On SonataMediaBundle I have another PR that solves the issue by providing a domain for the translator. But since there are no actual translations @core23 suggested that we don't need to translate the messages hence this PR.

OskarStark:
Thank you for the explanation, I mixed things up.
Can you please reword your changelog an commit message to be more clear why you need to do this change, and not what you change?

xaben:
Ok, reworded the Changelog message. Is it better now?

PR #212, opened 2015-11-23
[WIP] Add symfony 3.0 support

rande:

PR #412, opened 2017-07-18
Remove form compiler pass when mapping is disabled

deguif:
I am targeting this branch, because this is about about avoiding deprecation warnings.

## Changelog

```markdown
### Changed
- Removed add of `FormFactoryCompilerPass` when the option `form.mapping.enabled` is `false`
```

## Subject

I saw that there was an option for enabling form mapping (which defaults to true and was great for compatibility with Symfony <3)
As the `FormFactoryCompilerPass` was added  in #212 in the same time that this option, it seems that if we disable form mapping, the compiler pass is not necessary anymore.

Disabling the form mapping can be achieved with this config:
```yml
sonata_core:
    form:
        mapping:
            enabled: false
```

As the container parameter `sonata.core.form.mapping.type` is defined here ([DependencyInjection/SonataCoreExtension#L112](https://github.com/sonata-project/SonataCoreBundle/blob/3.x/DependencyInjection/SonataCoreExtension.php#L112)) based on the `form.mapping.enabled` configuration, I think we can use it to test if we need to add the `FormFactoryCompilerPass`

This will also avoid deprecation warnings for Symfony 4.0, but only when the form mapping is disabled. (The main purpose of this PR)

_I'm not sure if this is the right solution, but this is a proposal._

### DIFF
--- a/SonataCoreBundle.php
+++ b/SonataCoreBundle.php
@@ -27,7 +27,10 @@ class SonataCoreBundle extends Bundle
     {
         $container->addCompilerPass(new StatusRendererCompilerPass());
         $container->addCompilerPass(new AdapterCompilerPass());
-        $container->addCompilerPass(new FormFactoryCompilerPass());
+
+        if ($container->hasDefinition('sonata.core.form.extension.dependency')) {
+            $container->addCompilerPass(new FormFactoryCompilerPass());
+        }
 
         $this->registerFormMapping();
     }
--- a/Tests/SonataCoreBundleTest.php
+++ b/Tests/SonataCoreBundleTest.php
@@ -26,8 +26,13 @@ final class SonataCoreBundleTest extends PHPUnit_Framework_TestCase
     public function testBuild()
     {
         $containerBuilder = $this->getMockBuilder('Symfony\Component\DependencyInjection\ContainerBuilder')
-            ->setMethods(array('addCompilerPass'))
+            ->setMethods(array('addCompilerPass', 'hasDefinition'))
             ->getMock();
+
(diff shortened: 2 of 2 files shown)


greg0ire:
Or maybe test if `sonata.core.form.extension.dependency` still exists? It would be slightly more straightforward, right?

deguif:
As a complement to the PR description:

`SonataListFormMappingCommand` is relying on the `sonata.core.form.types` and `sonata.core.form.type_extensions` container parameters defined on this `FormFactoryCompilerPass`, but this can probably be avoided if the `form.mapping.enabled` is `false`

deguif:
@greg0ire Yeah sure I can test for `sonata.core.form.extension.dependency` service existence.

deguif:
Tests are failing because they are relying on the number of call to `addCompilerPass`. Will wait for more feedbacks before taking time to fix them, this would avoid losing my time if this PR is not suitable ;)

deguif:
Ok, just have some free time, I fixed the tests.

greg0ire:
Thanks @deguif !

Soullivaneuh:
@deguif PR reverted. Please see #435 for more information.

deguif:
Hi @Soullivaneuh 

Sorry for this bug introduction.
I will try to find another alternative to avoid the deprecation warnings.

Soullivaneuh:
@deguif This is not a deprecation warning, this produce a critical error.

deguif:
Yes @Soullivaneuh 
I just mean I now need to find another way to avoid the deprecation warning for Symfony 4 that my buggy previous PR was addressing.

PR #426, opened 2017-09-04
[WIP] Clean managers

core23:

## Changelog

```markdown
### Removed
 - Clean ManagerInterface by removing `getTableName`, `getConnection` and all `find*` methods
```

## To do

- [ ] Update implementations
- [ ] Add an upgrade note

## Subject

The managers are very technical and propagate some repository methods. Instead the concrete bundle manager should contain domain specific methods.

### DIFF
--- a/Model/ManagerInterface.php
+++ b/Model/ManagerInterface.php
@@ -11,8 +11,6 @@
 
 namespace Sonata\CoreBundle\Model;
 
-use Doctrine\DBAL\Connection;
-
 /**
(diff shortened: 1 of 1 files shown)


core23:
Not sure if we should remove the `find*` methods. The methods `getConnection` and `getTableName` should be removed considering ISP.

WDYT @greg0ire @jlamur

greg0ire:
If you are not sure then maybe move them to a different interface instead of removing them? I think interfaces should have from 1 to 5 methods ideally.

PR #429, opened 2017-09-15
Added missing datepicker options pickDate. Considered the possibility in the template to display only timepick.

Fuck4ik:


greg0ire:
Please associate your gmail email address with your Github account, or change the email in your commits to an address already associated with it.

Fuck4ik:
Done

composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "sonata-project/core-bundle",
+    "name": "omasn/core-bundle",
```
greg0ire: Why would you do that?

jlamur: Reverted in the last commit.

greg0ire:
Please get rid of the first and last commits by squashing or rebasing, then force pushing, also,

please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

Resources/views/Form/datepicker.html.twig
```diff
@@ -39,12 +39,14 @@ file that was distributed with this source code.
 
 {% block sonata_type_datetime_picker_widget_html %}
     {% if datepicker_use_button %}
-        <div class='input-group date' id='dtp_{{ id }}'>
+        <div class='input-group date {% if not dp_options['pickDate'] %}timepicker{% endif %}' id='dtp_{{ id }}'>
```
jlamur: @greg0ire WDYT of `if dp_options['pickDate'] is defined and not dp_options['pickDate']` for BC?

greg0ire: would be better indeed

Fuck4ik:
There are no cases when this option is absent

PR #430, opened 2017-09-15
Added missing datepicker options pickDate.

Fuck4ik:
Added the ability for datetime to display only time. The default icon now changes depending on the pickDate option

## Changelog

```markdown
### Added
- Added dp_option "dp_pick_date" in BasePickerType

### Changed
- Changed template datepicker.html.twig to display the time icon

### DIFF
--- a/Form/Type/BasePickerType.php
+++ b/Form/Type/BasePickerType.php
@@ -125,6 +125,7 @@ abstract class BasePickerType extends AbstractType
             'widget' => 'single_text',
             'datepicker_use_button' => true,
             'dp_pick_time' => true,
+            'dp_pick_date' => true,
             'dp_use_current' => true,
             'dp_min_date' => '1/1/1900',
             'dp_max_date' => null,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by greg0ire** [comment]
You could have just force pushed instead of creating a new PR...

**Comment by Fuck4ik** [comment]
Thanks, in the future I'll know it

File: Form/Type/BasePickerType.php [hunk 1]
```diff
@@ -125,6 +125,7 @@ protected function getCommonDefaults()
             'widget' => 'single_text',
             'datepicker_use_button' => true,
             'dp_pick_time' => true,
+            'dp_pick_date' => true,
```
greg0ire: @jlamur not sure there is a need for the expression you proposed if there is this default value.

Fuck4ik: pickDate = true, this is standard behavior sonata_type_datetime_picker. I just expanded the option to turn off date selection.

jlamur: I though about this hypothetical guy who extends this class, overrides the `getCommonDefaults()` method and doesn't call `parent::getCommonDefaults()` :thinking:

We will let this guy deal with this on its own. :)

greg0ire: Good thinking, but that's a bit like spacebar heating :P

greg0ire:

greg0ire:
Thanks @Fuck4ik

PR #437, opened 2017-09-29
Remove translation for sonata_template_box

core23:

## Changelog

```markdown
### Removed
 - Removed translation for `sonata_template_box`
```

## Subject

There is no good reason to translate the text which is only visibly to developers.

### DIFF
--- a/Twig/Extension/TemplateExtension.php
+++ b/Twig/Extension/TemplateExtension.php
@@ -13,7 +13,6 @@ namespace Sonata\CoreBundle\Twig\Extension;
 
 use Sonata\CoreBundle\Model\Adapter\AdapterInterface;
 use Sonata\CoreBundle\Twig\TokenParser\TemplateBoxTokenParser;
-use Symfony\Component\Translation\TranslatorInterface;
 
 class TemplateExtension extends \Twig_Extension
 {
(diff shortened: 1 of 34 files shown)


core23:
Please review @greg0ire @jlamur @OskarStark

Twig/Extension/TemplateExtension.php
```diff
@@ -28,19 +27,12 @@ class TemplateExtension extends \Twig_Extension
...
      */
-    protected $translator;
-
-    /**
-     * @param bool                $debug        Is Symfony debug enabled?
-     * @param TranslatorInterface $translator   Symfony Translator service
-     * @param AdapterInterface    $modelAdapter A Sonata model adapter
-     */
-    public function __construct($debug, TranslatorInterface $translator, AdapterInterface $modelAdapter)
+    public function __construct($debug, AdapterInterface $modelAdapter)
```
greg0ire: What about this BC-break? Do we consider it ok?

greg0ire: Oh wait this is for master, so ok.

greg0ire:
All these `__construct` changes are BC-breaks. How about taking this as an occasion to add typehinting to these signatures, and maybe also close the API (`final` ?)

Twig/Extension/TemplateExtension.php
core23: Don't know how to handle this without a big pain. So I created this for the master branch.

core23:
> All these `__construct` changes are BC-breaks. How about taking this as an occasion to add typehinting to these signatures, and maybe also close the API (`final` ?)

This should be done in an own PR which will close the complete API.

Twig/TokenParser/TemplateBoxTokenParser.php
```diff
@@ -55,7 +47,7 @@ public function parse(\Twig_Token $token)
 
         $this->parser->getStream()->expect(\Twig_Token::BLOCK_END_TYPE);
 
-        return new TemplateBoxNode($message, $translationBundle, $this->enabled, $this->translator, $token->getLine(), $this->getTag());
+        return new TemplateBoxNode($message, $translationBundle, $this->enabled, $token->getLine(), $this->getTag());
```
kunicmarko20: `$translationBundle` should have been removed here also.

PR #451, opened 2017-10-25
Restore php 5.6 through 7.0 compatibility

greg0ire:
## Changelog

```markdown
### Added
- php 5.6 - 7.0 compatibility
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ branches:
 language: php
 
 php:
+  - '5.6'
+  - '7.0'
   - '7.1'
   - nightly
 
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -10,6 +10,8 @@ branches:
 language: php
 
 php:
+  - '5.6'
+  - '7.0'
```
jlamur: Isn't this supposed to be managed by `dev-kit`?

greg0ire: It very much is, and will be ignored / overwritten. Can't make this PR without ensuring it actually works, right?

jlamur: Oh you'll create the PR on dev-kit after you have tested that no test fail, right?

greg0ire: Yeah, not sure what the plan is regarding the merge order though :/

jlamur: https://github.com/sonata-project/dev-kit/issues/315#issuecomment-339435597

PR #454, opened 2017-10-28
Register commands as services

ossinkine:
I am targeting this branch, because deprecation notices are thrown after update to Symfony 3.4.
Closes https://github.com/sonata-project/SonataAdminBundle/issues/4703

## Changelog

```markdown
### Fixed
- Register commands as services to prevent deprecation notices on Symfony 3.4
```

## Subject
After updating Symfony version to 3.4 the following deprecation notices are thrown:
> Auto-registration of the command "Sonata\CoreBundle\Command\SonataDumpDoctrineMetaCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead.

### DIFF
--- a/src/DependencyInjection/SonataCoreExtension.php
+++ b/src/DependencyInjection/SonataCoreExtension.php
@@ -78,6 +78,7 @@ EOT
         $loader->load('twig.xml');
         $loader->load('model_adapter.xml');
         $loader->load('core.xml');
+        $loader->load('commands.xml');
 
         $this->registerFlashTypes($container, $config);
         $container->setParameter('sonata.core.form_type', $config['form_type']);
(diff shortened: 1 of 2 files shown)


Resources/config/command.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="sonata.core.command.dump_doctrine_metadata" class="Sonata\CoreBundle\Command\SonataDumpDoctrineMetaCommand">
```
core23: No alias?

ossinkine: Which alias do you mean? Command name for tag or alias for service? And for what?

core23: Thought about [lazy command](https://symfony.com/blog/new-in-symfony-3-4-lazy-commands) aliases. But this is only possible in symfony 3.4+

OskarStark:
Thank you @ossinkine 👍

PR #456, opened 2017-11-02
Add legacy class to fix old Form compilation pass

nidble:
I am targeting this branch, because this does not break BC.

Closes sonata-project/SonataAdminBundle#4700
## Changelog
```
### Fixed
DependencyInjection/Compiler/FormFactoryCompilerPass.php
DependencyInjection/Compiler/LegacyFormFactoryCompilerPass.php
SonataCoreBundle.php
```
## Subject
Introduced legacy class to mantain compatibility with old sf versions and to provide a solution against deprecation warnings raised because old `FormFactoryCompilerPass` extends deprecated `FormPass`.

After this change if the new `FormPass` class is available, lt will be correctly extended by  `FormFactoryCompilerPass` while  `LegacyFormFactoryCompilerPass` still will mantain BC.

### DIFF
--- a/SonataCoreBundle.php
+++ b/SonataCoreBundle.php
@@ -11,11 +11,12 @@
 
 namespace Sonata\CoreBundle;
 
+use Sonata\CoreBundle\DependencyInjection\Compiler;
 use Sonata\CoreBundle\DependencyInjection\Compiler\AdapterCompilerPass;
-use Sonata\CoreBundle\DependencyInjection\Compiler\FormFactoryCompilerPass;
 use Sonata\CoreBundle\DependencyInjection\Compiler\StatusRendererCompilerPass;
 use Sonata\CoreBundle\Form\FormHelper;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\Form\DependencyInjection\FormPass;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 class SonataCoreBundle extends Bundle
@@ -27,7 +28,11 @@ class SonataCoreBundle extends Bundle
     {
         $container->addCompilerPass(new StatusRendererCompilerPass());
         $container->addCompilerPass(new AdapterCompilerPass());
-        $container->addCompilerPass(new FormFactoryCompilerPass());
+        if (class_exists(FormPass::class)) {
+            $container->addCompilerPass(new Compiler\FormFactoryCompilerPass());
+        } else {
+            $container->addCompilerPass(new Compiler\LegacyFormFactoryCompilerPass());
+        }
 
         $this->registerFormMapping();
     }
--- a/DependencyInjection/Compiler/LegacyFormFactoryCompilerPass.php
+++ b/DependencyInjection/Compiler/LegacyFormFactoryCompilerPass.php
@@ -11,10 +11,10 @@
 
 namespace Sonata\CoreBundle\DependencyInjection\Compiler;
 
+use Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\Form\DependencyInjection\FormPass;
 
-class FormFactoryCompilerPass extends FormPass
+class LegacyFormFactoryCompilerPass extends FormPass
 {
     /**
      * {@inheritdoc}


greg0ire:
Please write a changelog directed at end users. Keep in mind that they probably do not care what file or class you changed, but what features you provided / what bugs you fixed. Try to answer the question "What benefit do I get from upgrading?"
More on http://keepachangelog.com

DependencyInjection/Compiler/FormFactoryCompilerPass.php
```diff
@@ -40,17 +41,19 @@ public function process(ContainerBuilder $container)
         }
 
         // get factories
-        $original = $container->getDefinition('form.extension');
-        parent::process($container);
+        if (version_compare(Kernel::VERSION, '2.8', '<')) {
```
greg0ire: Please read this: https://symfony.com/doc/current/setup/bundles.html#updating-your-code-to-support-symfony-2-x-and-3-x-at-the-same-time

DependencyInjection/SonataCoreExtension.php
```diff
@@ -82,7 +82,7 @@ public function load(array $configs, ContainerBuilder $container)
         $this->registerFlashTypes($container, $config);
         $container->setParameter('sonata.core.form_type', $config['form_type']);
 
-        $this->configureFormFactory($container, $config);
+        // $this->configureFormFactory($container, $config);
```
greg0ire: Please remove commented code

greg0ire:
Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

greg0ire:
Can you explain what this does, and how you changed it? If I had to take a guess, I'd say you are disabling form mapping for all versions greater than 2.8 (or is it lower?) However, we do not support versions lower than 2.8 (look at composer.json), so these conditions are useless, right?

DependencyInjection/Compiler/FormFactoryCompilerPass.php
nidble: I used `version_compare Kernel::VERSION` because:
1) it's the first approach inside CompilerPass that I've seen in other CoreBundle files. 
2) this service is mainly used in the `SonataListFormMappingCommand` that through `isEnabled` checks the symfony version and not the presence of methods...

I think that better rewrite can be made further considering BC behavior from a global perspective...

greg0ire: > it's the first approach inside CompilerPass that I've seen in other CoreBundle files.

Mistakes have been made I'm afraid...

nidble: No problem dude :-) this is my first commit and I'm worrying about standard etc. that I've already messed.

thanks for your patience!

greg0ire: :) please read my last comment though, I think it's the most important: https://github.com/sonata-project/SonataCoreBundle/pull/456#issuecomment-341375779

nidble:
The hypothesis about my changes is to restrict the correct behavior only to certain sym version. i.e., considering the "client" class `SonataListFormMappingCommand`, this code must work with sym versions that differ from 3.x. The rest must be immutate.

To accomplish this I can fix my PR with `version_compare(Kernel::VERSION, '3, '!='), what do you think about?
thanks

greg0ire:
What do you mean by "sym"?

greg0ire:
Oh it's sf, ok

greg0ire:
I'm unsure if the limit should be 3.0 or 3.4, and also, I'm unsure what this means for versions that will not have the code inside the "if". Does it mean people will no longer be able to use `sonata_type_datetime_picker` inside `TheirAdmin::configureFormFields` ? If yes, that would be a big BC-break, right?

greg0ire:
I re read the error message and see that this is about extending another `FormPass` class. Since when does this `FormPass` class exist. IMO we should extend it as soon as it exists.

DependencyInjection/Compiler/FormFactoryCompilerPass.php
jlamur: BTW, we no longer support symfony versions < 2.8 🤔

greg0ire:
It exists since 3.3.0: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/CHANGELOG.md#330

nidble:
Regarding FormPass, actually there are two implementation, the old one is `Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass`(1). The new, introduced by me is: `Symfony\Component\Form\DependencyInjection\FormPass`(2).

The problem is that (1) raise deprecation warning. But if I try to remove build fail...

greg0ire:
A solution might be to have 2 passes too, and use the right one depending on the existence of `Symfony\Component\Form\DependencyInjection\FormPass` (no kernel version check please)

nidble:
Hi @greg0ire 
I worked on this PR and I made two different Pass. By checking existence of `Symfony\Component\Form\DependencyInjection\FormPass` now I can choose to add "new pass" or the oldest ` Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass`.

Before to proceed with a PR, whats will be the name of the new compile pass class?

greg0ire:
I'd rather you keep the old name and rename the old compiler pass to `LegacyWhateverTheCurrentNameIs`

SonataCoreBundle.php
```diff
@@ -27,7 +28,12 @@ public function build(ContainerBuilder $container)
     {
         $container->addCompilerPass(new StatusRendererCompilerPass());
         $container->addCompilerPass(new AdapterCompilerPass());
-        $container->addCompilerPass(new FormFactoryCompilerPass());
+        if (class_exists(FormPass::class)) {
+            $container->addCompilerPass(new Compiler\FormFactoryCompilerPass());
+        }
+        else {
+            $container->addCompilerPass(new Compiler\LegacyFormFactoryCompilerPass());
+        }
```
greg0ire: This will make StyleCi angry, please use PSR-2. Also, you could use a ternary inside `addCompilerPass` instead.

greg0ire:
This should help fixing StyleCi:

```
git fetch --all --prune
git rebase origin/3.x
git push --force
```

Hoping origin is the main repo here.

nidble:
> git fetch --all --prune

Fetching origin
> git rebase origin/3.x

Current branch 3.x is up to date.
> git push --force

Everything up-to-date

greg0ire:
`git remote -vvv`?

nidble:
> git remote -vvv

origin	https://github.com/nidble/SonataCoreBundle.git (fetch)
origin	https://github.com/nidble/SonataCoreBundle.git (push)

greg0ire:
- `git remote rename origin fork`
- `git remote add origin https://github.com/sonata-project/SonataCoreBundle.git`

and then try again previous steps

greg0ire:
StyleCI is green, good job :)

greg0ire:
Travis is failing though, please fix it.

greg0ire:
No response... I'm working on this myself now.

greg0ire:
Ok so here is my understanding of the situation: the sonata core compiler pass listens to the symfony compiler pass, copies the names of all forms and gives it to https://github.com/sonata-project/SonataCoreBundle/blob/3.x/Form/Extension/DependencyInjectionExtension.php#L20-L23 , which allows to use forms "like before". This will probably not work in Symfony 4, and we should stop supporting it.

So here what I think should be done:
- add an option to disable this;
- trigger a deprecation error if the user does not disable it.

greg0ire:
Looks like that setting already exists: https://github.com/sonata-project/SonataCoreBundle/blob/3.x/DependencyInjection/Configuration.php#L45

nidble:
@greg0ire 
sorry but before to answer you, it was more accurate to fully understand if was necessary or not to provide two different test case...

But now, what did you think that default configuration would be `false`? And what's the best code strategy to make test pass and code done right?

greg0ire:
If you want this make this work, you should find a way to get the array of form type names / form type extensions from the new `FormPass` (the one from dependency injection, as opposed from the old one, from framework bundle). I haven't looked into it, because I believe people should no longer use this feature. Working on a PR to deprecate it right now.

> what did you think that default configuration would be false?

That would be a huge BC break, so definitely not. We should disable this feature if the user uses Symfony 4, BTW.

greg0ire:
See my tweet for how to get rid of the the deprecation warning: https://twitter.com/greg0ire/status/927256894724608000

greg0ire:
Also see #462

nidble:
Make the appropriate changes to this PR, imho involve in two main point:
- improve compilerpass to handle `sonata.core.form.extension.dependency` service
- make a proper test case

But doing this, for a service that is used only by `SonataListFormMappingCommand` when sf !== 3, with the restriction of your PR, for me is a little bit expensive...

greg0ire:
@nidble you do as you feel, if it's too expensive, don't do it: as I said, people can get rid of the deprecation error the hard (but correct way):

```yaml
sonata_core: 
    form: 
        mapping: 
            enabled: false
```

Well, at least I hope! It would be great if someone actually tried this on their application and reported back 😅

nidble:
Sure, previously with sf !== 3, I refered at the condition inside `SonataListFormMappingCommand` that enable this command, the only one class involved in service utilization.

Providing new version that works in all circumstances is not useful if that command only works for before condition https://github.com/sonata-project/SonataCoreBundle/blob/3.x/src/Command/SonataListFormMappingCommand.php#L32

greg0ire:
Ok, let's close this then?

PR #457, opened 2017-11-04
Sf 3.4 compat

greg0ire:
I am targeting this branch, because this does not break BC.
## Changelog

```markdown
### Fixed
- reusable test case `Sonata\CoreBundle\Test\AbstractWidgetTestCase` is now compatible with Symfony 3.4
```

## Subject

This fixes the 3.4 build.

### DIFF
--- a/Test/AbstractWidgetTestCase.php
+++ b/Test/AbstractWidgetTestCase.php
@@ -16,9 +16,11 @@ use Symfony\Bridge\Twig\Extension\TranslationExtension;
 use Symfony\Bridge\Twig\Form\TwigRenderer;
 use Symfony\Bridge\Twig\Form\TwigRendererEngine;
 use Symfony\Bridge\Twig\Form\TwigRendererEngineInterface;
+use Symfony\Bridge\Twig\Form\TwigRendererInterface;
 use Symfony\Bridge\Twig\Tests\Extension\Fixtures\StubFilesystemLoader;
 use Symfony\Bundle\FrameworkBundle\Tests\Templating\Helper\Fixtures\StubTranslator;
 use Symfony\Component\Form\FormExtensionInterface;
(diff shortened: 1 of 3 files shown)


Test/AbstractWidgetTestCase.php
```diff
@@ -88,6 +78,11 @@ protected function setUp()
         $this->extension->initRuntime($environment);
     }
 
+    public function getRenderer()
```
Soullivaneuh: Missing PHPDoc.

Test/AbstractWidgetTestCase.php
```diff
@@ -89,6 +80,14 @@ protected function setUp()
     }
 
     /**
+     * @return TwigRendererInterface
+     */
+    public function getRenderer()
```
core23: final?

greg0ire: Sure. Done!

Test/AbstractWidgetTestCase.php
```diff
@@ -50,36 +53,24 @@ protected function setUp()
...
             );
-            $runtimeLoader = $this
-                ->getMockBuilder('Twig_RuntimeLoaderInterface')
-                ->getMock();
-
-            $runtimeLoader->expects($this->any())
-                ->method('load')
-                ->with($this->equalTo('Symfony\Bridge\Twig\Form\TwigRenderer'))
-                ->will($this->returnValue($this->renderer));
+            $runtimeLoader = new \Twig_FactoryRuntimeLoader([
```
Soullivaneuh: You can use this: https://github.com/twigphp/Twig/blob/1.x/src/RuntimeLoader/FactoryRuntimeLoader.php

Twig 1.34 is required.

Soullivaneuh: Let's do it later.

PR #469, opened 2017-11-17
Namespaced twig

greg0ire:

## Changelog

```markdown
### Fixed
- `twig_truncate_filter`, and `twig_wordwrap_filter` were broken
- `twig_truncate_filter`, and `twig_wordwrap_filter` were not properly deprecated
```

## Subject

The goal of this PR was to use Twig namespaces, but I found 2 bugs along the way.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "symfony/translation": "^2.8 || ^3.2",
         "symfony/twig-bridge": "^2.8 || ^3.2",
         "symfony/validator": "^2.8 || ^3.2",
-        "twig/extensions": "^1.0",
+        "twig/extensions": "^1.5",
         "twig/twig": "^1.34 || ^2.0"
     },
     "require-dev": {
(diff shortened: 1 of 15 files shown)


composer.json
```diff
@@ -43,7 +43,7 @@
         "sensio/framework-extra-bundle": "^3.0 || ^4.0 || ^5.0",
         "sllh/php-cs-fixer-styleci-bridge": "^2.0",
         "sonata-project/exporter": "^1.3",
-        "symfony/phpunit-bridge": "^3.3"
+        "symfony/phpunit-bridge": "^3.3.12"
```
jordisala1991: We will need this change on all bundles to use SymfonyTestListener?

greg0ire: Yes, otherwise you get a signature mismatch issue. Not a big deal since this is a dev dependency.

phpunit.xml.dist
```diff
@@ -17,6 +17,10 @@
         </testsuite>
     </testsuites>
 
+    <listeners>
```
OskarStark: Couldn't we manage this file through dev-it?

greg0ire: That would be a good idea I think. Now that we have a standard structure.

greg0ire:

PR #478, opened 2017-11-26
Manual merge

greg0ire:
This merges the huge d45f0d2 (so the merge commit + some subsequent commits)

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,7 +19,7 @@
 
     <filter>
         <whitelist>
-            <directory suffix="Test.php">./src</directory>
+            <directory suffix=".php">./src/</directory>
         </whitelist>
     </filter>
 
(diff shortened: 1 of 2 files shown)


phpunit.xml.dist
```diff
@@ -18,8 +18,8 @@
     </testsuites>
 
     <filter>
-        <ini name="precision" value="8"/><whitelist>
```
greg0ire: I introduced this weird tag in the previous PR, sorry.

phpunit.xml.dist
```diff
@@ -18,8 +18,8 @@
...
-            <directory suffix="Test.php">./src</directory>
+        <whitelist>
+            <directory suffix=".php">./src/</directory>
```
greg0ire: And here I put an unwanted `Test`...

PR #471, opened 2017-11-17
Fw compat

greg0ire:
I am targeting this branch, because this is BC.

Closes https://github.com/sonata-project/SonataAdminBundle/issues/4700

## Changelog

```markdown
### Fixed
- forward-compatibility with sf4
```

PR #490, opened 2017-12-05
Avoid inheritance from deprecated FormPass

ossinkine:
I am targeting this branch, because after many labors we did not manage to get rid of the deprecation message.

Relates #489 #471 #412 sonata-project/SonataAdminBundle#4700

## Changelog

```markdown
### Changed
- Avoid inheritance from deprecated FormPass
```

## To do

- [x] Update the tests

### DIFF
--- a/src/DependencyInjection/Compiler/FormFactoryCompilerPass.php
+++ b/src/DependencyInjection/Compiler/FormFactoryCompilerPass.php
@@ -12,12 +12,13 @@
 namespace Sonata\CoreBundle\DependencyInjection\Compiler;
 
 use Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
  * @deprecated since 3.7, to be removed in 4.0, the form mapping feature should be disabled.
  */
-class FormFactoryCompilerPass extends FormPass
+class FormFactoryCompilerPass implements CompilerPassInterface
 {
     /**
      * {@inheritdoc}
@@ -44,7 +45,7 @@ class FormFactoryCompilerPass extends FormPass
 
         // get factories
         $original = $container->getDefinition('form.extension');
-        parent::process($container);
+        $this->processFormPass($container);
 
         $factory = $container->getDefinition('sonata.core.form.extension.dependency');
         $factory->replaceArgument(1, $original->getArgument(1));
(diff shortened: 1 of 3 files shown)


greg0ire:
Great idea with the composition! Please go on!

ossinkine:
Please review

src/DependencyInjection/Compiler/FormFactoryCompilerPass.php
```diff
@@ -12,18 +12,30 @@
...
 
 use Symfony\Bundle\FrameworkBundle\DependencyInjection\Compiler\FormPass;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
  * @deprecated since 3.7, to be removed in 4.0, the form mapping feature should be disabled.
  */
-class FormFactoryCompilerPass extends FormPass
+class FormFactoryCompilerPass implements CompilerPassInterface
```
greg0ire: I think this is an acceptable BC-break

jordisala1991: Not sure if this is a BC-break: FormPass implements CompilerPassInterface. Isn't it allowed in this case?

greg0ire: It's a BC-break for code that does `instanceof FormPass`. So, acceptable, allowed, whatever adjective floats your boat.

greg0ire:
Why is this WIP now?

ossinkine:
It seems the form mapping check is useless, because SonataCoreExtension already checks it and removes `sonata.core.form.extension.dependency` definition. So replace inheritance with composition without this checking is enough.

src/DependencyInjection/Compiler/FormFactoryCompilerPass.php
```diff
@@ -56,4 +57,10 @@ public function process(ContainerBuilder $container)
 
         $container->setDefinition('form.extension', $factory);
     }
+
+    protected function processFormPass(ContainerBuilder $container)
```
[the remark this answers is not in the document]
ossinkine: Do you mean it should be private? What if anyone wants to override it?

ossinkine:
It works for me in Symfony 3.4 but I can't test it on Symfony 2.8, can anyone do it?

src/DependencyInjection/Compiler/FormFactoryCompilerPass.php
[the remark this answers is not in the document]
greg0ire: They will tell us, and we will see if we should open it or modify it http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html

Xymanek:
Just tested on 3.4 and it the deprecation is gone 👍

greg0ire:
@ossinkine the form pass isn't triggering a deprecation on 2.8 anyway right?

ossinkine:
@greg0ire Yes, but I will be more relaxed if anyone ensures that all works on 2.8

greg0ire:
@sonata-project/contributors please merge if we don't have 2.8 feedback this evening

PR #491, opened 2017-12-06
Mark form type services as deprecated

ossinkine:
I am targeting this branch, because deprecated classes trigger notification if according services are not marked as deprecated.

## Changelog

```markdown
### Fixed
- Fixed deprecation message in `ColorSelectorType`
```

### DIFF
--- a/src/Form/Type/TranslatableChoiceType.php
+++ b/src/Form/Type/TranslatableChoiceType.php
@@ -19,6 +19,16 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
+@trigger_error(
+    sprintf(
+        'Form type "%s" is deprecated since SonataCoreBundle 2.2.0 and will be'
+        .' removed in 4.0. Use form type "%s" with "translation_domain" option instead.',
+        TranslatableChoiceType::class,
+        ChoiceType::class
+    ),
+    E_USER_DEPRECATED
+);
+
 /**
  * NEXT_MAJOR: remove this class.
  *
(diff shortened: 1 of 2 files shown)


greg0ire:
Blocked by the xmllint issue

ossinkine:
Rebased

src/Form/Type/TranslatableChoiceType.php
```diff
@@ -37,12 +43,6 @@ class TranslatableChoiceType extends AbstractType
      */
     public function __construct(TranslatorInterface $translator)
     {
-        @trigger_error(
```
jordisala1991: Why do we need to move this `trigger_error`?

ossinkine: It has been moved to head of the file because error should be thrown on file inclusion but not on object creation

jordisala1991: I am not sure of this change, because we are not super consistent with where we put those `trigger_error`. Normally we do it in the `__construct`. Would like to hear the opinion of @greg0ire on this also

greg0ire: I've seen it in the header too (Admin.php)

src/Form/Type/TranslatableChoiceType.php
```diff
@@ -19,6 +19,12 @@
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
+@trigger_error(
+    'Form type "sonata_type_translatable_choice" is deprecated since SonataCoreBundle 2.2.0 and will be'
```
greg0ire: Use the class FQCN instead please

src/Form/Type/TranslatableChoiceType.php
```diff
@@ -19,6 +19,12 @@
...
+    .' removed in 4.0. Use form type "choice" with "translation_domain" option instead.',
```
greg0ire: Same here

src/Form/Type/TranslatableChoiceType.php
ossinkine: Class inheritance also should trigger error, so function call should be placed in header.

PR #495, opened 2017-12-08
Add twig extension for deprecating templates

kunicmarko20:
I am targeting this branch, because I am adding a twig function for deprecating templates, needed for https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/772 and https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/258

## Changelog

```markdown
### Added
- Added twig extension for deprecating templates
```

## To do

- [x] Add tests
- [x] Update documentation
## Subject

We needed a way to deprecate old templates, this is used for https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/772 and https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/pull/258 as suggested it is better to be in this bundle than in `SonataAdminBundle`.

### DIFF
--- a/src/Twig/Node/DeprecatedTemplateNode.php
+++ b/src/Twig/Node/DeprecatedTemplateNode.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Twig\Node;
+
+use Twig\Compiler;
+use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Node;
+
+/**
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+final class DeprecatedTemplateNode extends Node
+{
+    public function __construct(AbstractExpression $newTemplate, $line, $tag = null)
+    {
+        parent::__construct(['newTemplate' => $newTemplate], [], $line, $tag);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function compile(Compiler $compiler)
+    {
+        @trigger_error(sprintf(
+            'The "%s" template is deprecated. Use "%s" instead.',
+            $this->getTemplateName(),
+            $this->getNode('newTemplate')->getAttribute('value')
+        ), E_USER_DEPRECATED);
+    }
+}
--- a/src/Twig/TokenParser/DeprecatedTemplateTokenParser.php
+++ b/src/Twig/TokenParser/DeprecatedTemplateTokenParser.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Twig\TokenParser;
+
+use Sonata\CoreBundle\Twig\Node\DeprecatedTemplateNode;
+use Twig\Token;
+use Twig\TokenParser\AbstractTokenParser;
+
+/**
+ * @author Marko Kunic <kunicmarko20@gmail.com>
+ */
+final class DeprecatedTemplateTokenParser extends AbstractTokenParser
+{
+    public function parse(Token $token)
+    {
+        if (!$this->parser->getStream()->test(Token::STRING_TYPE)) {
+            throw new \InvalidArgumentException('New template name is mandatory.');
+        }
+
+        $newTemplate = $this->parser->getExpressionParser()->parseExpression();
+
+        $this->parser->getStream()->expect(Token::BLOCK_END_TYPE);
+
+        return new DeprecatedTemplateNode($newTemplate, $token->getLine(), $this->getTag());
+    }
+
+    public function getTag()
+    {
+        return 'sonata_template_deprecate';
+    }
+}
--- a/tests/Twig/Node/DeprecatedTemplateNodeTest.php
+++ b/tests/Twig/Node/DeprecatedTemplateNodeTest.php
@@ -1,0 +1,58 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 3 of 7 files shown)


core23:
Would prefer a custom node like:
```twig
{% sonata_template_deprecate 'Acme:NewTemplate' %}
```

kunicmarko20:
Well, would be nice if I knew this before I moved it for the second time... Do rest of people agree with Node instead of Function?

jordisala1991:
... well I think it is better a node, but I dont really care if it is a function. Do we really need to implement a node? wdyt @greg0ire ?

OskarStark:
node looks indeed better, sorry for the inconvenience @kunicmarko20 🍺

greg0ire:
A node? What's a node? If it translates to a node in the DOM, then it makes little sense, does it? A function is just fine IMO.

@kunicmarko20 very sorry for all the inconveniences 😅 , wait until we have a clear consensus.

core23:
We already use nodes, e.g. `sonata_template_box`.

This are custom tags which extends the normale twig syntax

greg0ire:
To me, node or tags have somehow to do with content. This doesn't, that's why it feels weird to me.

jvasseur:
@greg0ire to me it's the opposite, function are used to generate content by returning a value. Tags are language constructs that can be used for things that doesn't fit the function use case.

greg0ire:
Makes sense. Then let's go with a tag if it's more natural for both of you.

kunicmarko20:
Changed to node :tada:, if it looks ok and that is what you wanted I will add tests.

src/Twig/Node/DeprecatedTemplateNode.php
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function compile(Compiler $compiler)
+    {
+        @trigger_error(sprintf(
+            'The "%s" template is deprecated. Use "%s" instead.',
+            $this->getNode('oldTemplate')->getAttribute('value'),
```
core23: Do we need the `oldTemplate`? Can we resolve this automatically with `$this->getTemplateName()` like in
https://github.com/sonata-project/SonataCoreBundle/blob/11e833fbcf04a108b6b1606344ffa2819505a15f/src/Twig/Node/TemplateBoxNode.php#L82?

kunicmarko20: @core23 removed, will add tests tomorrow, where should I add documentation for this?

src/Twig/Node/DeprecatedTemplateNode.php
```diff
@@ -0,0 +1,36 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Twig\Node;
+
+use Twig\Compiler;
+use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Node;
+
+class DeprecatedTemplateNode extends Node
```
OskarStark: please add `@author` tag

src/Twig/TokenParser/DeprecatedTemplateTokenParser.php
```diff
@@ -0,0 +1,37 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\CoreBundle\Twig\TokenParser;
+
+use Sonata\CoreBundle\Twig\Node\DeprecatedTemplateNode;
+use Twig\Token;
+use Twig\TokenParser\AbstractTokenParser;
+
+class DeprecatedTemplateTokenParser extends AbstractTokenParser
```
OskarStark: please add `@author` tag

tests/Twig/Node/DeprecatedTemplateNodeTest.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Sonata\CoreBundle\Tests\Twig\Node;
+
+use Twig\Test\NodeTestCase;
+use Sonata\CoreBundle\Twig\Node\DeprecatedTemplateNode;
+use Twig\Node\Expression\ConstantExpression;
+
+class DeprecatedTemplateNodeTest extends NodeTestCase
```
OskarStark: same here

tests/Twig/TokenParser/DeprecatedTemplateTokenParserTest.php
```diff
@@ -0,0 +1,57 @@
...
+use PHPUnit\Framework\TestCase;
+use Sonata\CoreBundle\Twig\Node\DeprecatedTemplateNode;
+use Sonata\CoreBundle\Twig\TokenParser\DeprecatedTemplateTokenParser;
+use Twig\Source;
+use Twig\Environment;
+use Twig\Loader\ArrayLoader;
+use Twig\Parser;
+use Twig\Node\Expression\ConstantExpression;
+
+class DeprecatedTemplateTokenParserTest extends TestCase
```
OskarStark: and here

src/Twig/Node/DeprecatedTemplateNode.php
core23: Please make this final

src/Twig/TokenParser/DeprecatedTemplateTokenParser.php
core23: final

kunicmarko20:
@core23 Where should I add documentation for this? As this is not something developers will use, but only contributors. Everything else is fixed.

greg0ire:
> @core23 Where should I add documentation for this? As this is not something developers will use, but only contributors. Everything else is fixed.

How about creating a new document?

kunicmarko20:
@greg0ire you mean like a new page in Core documentation, sounds fine to me, do others agree?

greg0ire:
Yes I mean that. I don't think there is some kind of internal docs right now

kunicmarko20:
I can't find it so I do not think there is something like that, I will add it tonight, is there maybe anything else I should document in there besides this feature?

greg0ire:
> I can't find it so I do not think there is something like that, I will add it tonight, is there maybe anything else I should document in there besides this feature?

Can't think of anything right now :thinking: If you think the page will be maybe too small, then embed it as a paragraph somewhere else…

kunicmarko20:
Well, it will be too small, the only place I maybe see it fit is https://sonata-project.org/bundles/core/3-x/doc/reference/twig_helpers.html as it is a twig helper.

kunicmarko20:
screenshot from 2017-12-12 21-18-12

Documentation is added, can you now review everything one more time?

OskarStark:
Congratulations, you did it, thank you for your patience 🎉 👏

kunicmarko20:
It was my pleasure and thank you all for helping me.

PR #499, opened 2017-12-10
Always flip choices in BaseStatusType

core23:
I am targeting this branch, because this is a bugfix, but also a soft BC break.

## Changelog

```markdown
### Changed
 - Always flip choices in `BaseStatusType`
```

## Subject

The commit https://github.com/sonata-project/SonataCoreBundle/commit/1eefe7607bbf20ad5ef401febbde6110fb9269f2#diff-c25f4304e73e2ac2efcf019f17558604 introduced a wrong approach which will cause some problems when using symfony 3.

### DIFF
--- a/src/Form/Type/BaseStatusType.php
+++ b/src/Form/Type/BaseStatusType.php
@@ -13,7 +13,6 @@ namespace Sonata\CoreBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
-use Symfony\Component\Form\FormTypeInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
@@ -104,7 +103,7 @@ abstract class BaseStatusType extends AbstractType
         $choices = call_user_func([$this->class, $this->getter]);
 
         // choice_as_value options is not needed in SF 3.0+
-        if (method_exists(FormTypeInterface::class, 'setDefaultOptions')) {
+        if ($resolver->isDefined('choices_as_values')) {
             $resolver->setDefault('choices_as_values', true);
         }
 
--- a/src/Form/Type/BaseStatusType.php
+++ b/src/Form/Type/BaseStatusType.php
@@ -13,6 +13,7 @@ namespace Sonata\CoreBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\Form\FormTypeInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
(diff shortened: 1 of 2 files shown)


tests/Form/Type/StatusTypeTest.php
```diff
@@ -60,6 +60,29 @@ public function testGetParent()
...
+    /**
+     * @group legacy
+     */
+    public function testGetDefaultOptionsLegacy()
+    {
+        Choice::$list = [
+            1 => 'salut',
+        ];
+
+        $type = new StatusType('Sonata\CoreBundle\Tests\Form\Type\Choice', 'getList', 'choice_type', true);
```
OskarStark: Please use `::class constant here

tests/Form/Type/StatusTypeTest.php
```diff
@@ -87,7 +110,7 @@ public function testGetDefaultOptionsWithValidFlip()
             2 => 'toi!',
         ];
 
-        $type = new StatusType('Sonata\CoreBundle\Tests\Form\Type\Choice', 'getList', 'choice_type', true);
+        $type = new StatusType('Sonata\CoreBundle\Tests\Form\Type\Choice', 'getList', 'choice_type');
```
OskarStark: same here

tests/Form/Type/StatusTypeTest.php
```diff
@@ -109,7 +132,7 @@ public function testGetDefaultOptionsWithValidInvalidFlip()
             2 => 'error',
         ];
 
-        $type = new StatusType('Sonata\CoreBundle\Tests\Form\Type\Choice', 'getList', 'choice_type', true);
+        $type = new StatusType('Sonata\CoreBundle\Tests\Form\Type\Choice', 'getList', 'choice_type');
```
OskarStark: and here

core23:
Please review again @OskarStark @greg0ire

src/Form/Type/BaseStatusType.php
```diff
@@ -93,6 +103,12 @@ public function configureOptions(OptionsResolver $resolver)
     {
         $choices = call_user_func([$this->class, $this->getter]);
 
+        // choice_as_value options is not needed in SF 3.0+
+        if (method_exists(FormTypeInterface::class, 'setDefaultOptions')) {
```
greg0ire: I think @EmmanuelVella has a better way for this, `$resolver->defined('choices_as_value') ` iirc

src/Form/Type/BaseStatusType.php
```diff
@@ -93,6 +103,12 @@ public function configureOptions(OptionsResolver $resolver)
     {
         $choices = call_user_func([$this->class, $this->getter]);
 
+        // choice_as_value options is not needed in SF 3.0+
+        if ($resolver->defined('choices_as_values')) {
```
EmmanuelVella: It should be `isDefined ;)`

jordisala1991:
The expected behavior for this type is to pass the choices without the flip applied? This seems a little wrong to use. This would mean that I have to do it like the old way and expect this type to flip the choices, and if I want to do it the new way, then I get a deprecated message.

@core23  why the default value is not false?

core23:
> The expected behavior for this type is to pass the choices without the flip applied?

That's right, but we need this old behavior because of BC reasons. A better way might deprecating this class and create a new one which don't need the flip anymore.

PR #500, opened 2017-12-10
Remove tautological comments

greg0ire:


src/Model/BasePHPCRManager.php
```diff
@@ -18,8 +18,6 @@
     /**
      * Make sure the code is compatible with legacy code.
      *
-     * @param $name
-     *
      * @return mixed
      */
```
jlamur: @greg0ire Two possibilities here: you add `@return mixed` and you type hint `@param mixed $var` everywhere when no php type hint is available ; or you remove all `mixed` docs type hints. This was not voted nor discussed IIRC.

greg0ire: @jlamur ^^ no type hint does not mean it is necessarily mixed. Here, we know that at least 2 types can be returned.

jordisala1991:
Thank you @greg0ire

PR #521, opened 2018-02-13
Update templates paths in docs

covex-nn:
I am targeting this branch, because this is docs patch.

## Subject

This PR updates templates paths to .twig files in docs, using twig namespaced syntax instead of templating syntax


jordisala1991:
Thank you @covex-nn

PR #522, opened 2018-02-13
Minor composer improvements

jordisala1991:

## Subject

Doing some archeology here:

- Removed FOSRestBundle and Sensio Framework Extra Bundle. They were added here: https://github.com/sonata-project/SonataCoreBundle/pull/45 but that code was removed here: https://github.com/sonata-project/SonataCoreBundle/commit/e781255d0b6ab661f9e041b01ff8d943ae70ebc0#diff-fdd764ee72f777e180ace83412a44158 without removing the dependencies. I cannot find any mention, and the tests does not show any difference without having those bundles installed.

- Allow newer versions of `serializer-bundle`, `symfony-config-test` and `matthiasnoback/symfony-dependency-injection-test`. This will allow us to add Symfony 4.0 to travis. With this changes all packages here are compatible with SF 4.0

- Upgrade PHPUnit bridge

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,13 +40,11 @@
     "require-dev": {
         "doctrine/orm": "^2.4",
         "doctrine/phpcr-odm": "^1.0",
-        "friendsofsymfony/rest-bundle": "^1.1 || ^2.0",
         "jackalope/jackalope-doctrine-dbal": "^1.0",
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @jordisala1991 !

PR #523, opened 2018-02-14
Make services public

covex-nn:
I am targeting this branch, because this is patch.

## Changelog

```markdown
### Changed
- Services for slugify strings are marked as public
```
## Subject

Services `sonata.core.slugify.cocur` and `sonata.core.slugify.native` can be used in SonataPageBundle via container and must be public. See `Sonata\PageBundle\SonataPageBundle::boot`


greg0ire:
Thanks @covex-nn !

PR #536, opened 2018-04-19
prepare for next major

kunicmarko20:
## Subject

Removing not needed stuff.


kunicmarko20:
Can you block so someone doesn't merge, just realized I did not check typehint/returntypes

greg0ire:
I think that this should be a separate PR

kunicmarko20:
oh, ok then. Will do it after this is merged.

src/SonataCoreBundle.php
```diff
@@ -146,10 +139,6 @@ public function registerFormMapping(): void
             'form.type_extension.form.data_collector',
         ];
 
-        if (class_exists(DescriptionFormTypeExtension::class)) {
-            $formTypes[] = 'nelmio_api_doc.form.extension.description_form_type_extension';
-        }
-
         FormHelper::registerFormExtensionMapping('form', $formTypes);
```
jordisala1991: I think everything related to FormHelper should get removed

kunicmarko20: There wasn't any kind of comment so I wasn't sure :confused:

jordisala1991: @greg0ire the formHelper is used only to keep compat with old form types declaration, or has more uses?

greg0ire: Not sure… maybe try removing it in a real app with form mapping disabled?

kunicmarko20: found the depreciation, will remove it now.

kunicmarko20:
should be :ok_hand: now

jordisala1991:
Do we need to modify the upgrade notes or something?

kunicmarko20:
I don't think so, I think @Soullivaneuh does that when doing a releas, everything else should be already noted when deprecated?

greg0ire:
Great job @kunicmarko20 !

PR #538, opened 2018-05-17
Minor Date(Time)PickerType integration improvement

Kal74:

```markdown
### Added
- [Date(Time)PickerType] Handle field without seconds with option dp_use_seconds.
- [Date(Time)PickerType] Allow DateTime Object in dp_min_date and dp_max_date
- [Date(Time)RangePickerType] Link DateTimeRange pickers to keep start < end
```

## Subject
Small update on the Date(Time)PickerType to 
- allow define field without seconds, and keep Intl localization (use IntlDateFormatter::SHORT).
- usage of Php \DateTime Object instead of string in the min & max values
- Range type are now linked together like described https://eonasdan.github.io/bootstrap-datetimepicker/#linked-pickers

### DIFF
--- a/src/Form/Type/BasePickerType.php
+++ b/src/Form/Type/BasePickerType.php
@@ -72,7 +72,7 @@ abstract class BasePickerType extends AbstractType
         if (isset($options['date_format']) && is_string($options['date_format'])) {
             $format = $options['date_format'];
         } elseif (is_int($format)) {
-            $timeFormat = ($options['dp_pick_time']) ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::NONE;
+            $timeFormat = ($options['dp_pick_time']) ? ($options['dp_use_seconds'] ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::SHORT) : \IntlDateFormatter::NONE;
             $intlDateFormatter = new \IntlDateFormatter(
                 \Locale::getDefault(),
                 $format,
--- a/src/Form/Type/BasePickerType.php
+++ b/src/Form/Type/BasePickerType.php
@@ -72,9 +72,12 @@ abstract class BasePickerType extends AbstractType
         if (isset($options['date_format']) && is_string($options['date_format'])) {
             $format = $options['date_format'];
         } elseif (is_int($format)) {
-            $timeFormat = ($options['dp_pick_time']) ? ($options['dp_use_seconds'] ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::SHORT) : \IntlDateFormatter::NONE;
+            $timeFormat = \IntlDateFormatter::NONE;
+            if ($options['dp_pick_time']) {
+                $timeFormat = $options['dp_use_seconds'] ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::SHORT;
+            }
             $intlDateFormatter = new \IntlDateFormatter(
-                \Locale::getDefault(),
+                $this->locale,
                 $format,
                 $timeFormat,
                 null,
--- a/src/Form/Type/BasePickerType.php
+++ b/src/Form/Type/BasePickerType.php
@@ -87,6 +87,13 @@ abstract class BasePickerType extends AbstractType
         // use seconds if it's allowed in format
         $options['dp_use_seconds'] = false !== strpos($format, 's');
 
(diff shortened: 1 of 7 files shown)


src/Form/Type/BasePickerType.php
```diff
@@ -72,7 +72,7 @@ public function finishView(FormView $view, FormInterface $form, array $options)
         if (isset($options['date_format']) && is_string($options['date_format'])) {
             $format = $options['date_format'];
         } elseif (is_int($format)) {
-            $timeFormat = ($options['dp_pick_time']) ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::NONE;
+            $timeFormat = ($options['dp_pick_time']) ? ($options['dp_use_seconds'] ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::SHORT) : \IntlDateFormatter::NONE;
```
kunicmarko20: This is confusing, I think 1 ternary should be the limit.

src/Resources/views/Form/datepicker.html.twig
```diff
@@ -67,3 +85,21 @@ file that was distributed with this source code.
         </script>
     {% endspaceless %}
 {% endblock sonata_type_datetime_picker_widget %}
+
+{% block sonata_type_datetime_range_picker_widget %}
+    {% spaceless %}
+    {{ block('form_widget') }}
+        <script type="text/javascript">
```
kunicmarko20: This block looks like a copy paste from the other one, extract it and reuse. Are you sure we dont have this block somewhere already?

OskarStark:
Could you please add some tests?

src/Resources/views/Form/datepicker.html.twig
Kal74: all block in this file are duplicated from date_picker and date_time picker just for one char in the id: dtp_xxx vs dp_xxx.
I wanted to keep it in the same style, but you're right

src/Form/Type/BasePickerType.php
Kal74: is it just a syntax problem, or do you dislike the idea of using the dp_use_seconds option?
I imagined otherwise to add an option time_format which would be by default DateTimeType::DEFAULT_TIME_FORMAT but that we could change to \IntlDateFormatter::SHORT.
I didn't did it because of the dp_use_seconds option which was defined, but which did not work as is.

src/Resources/views/Form/datepicker.html.twig
greg0ire: Bonus points for a separate js files that just parses the DOM and extracts needed data from data-attributes (no js in Twig would be great)

src/Form/Type/BasePickerType.php
kunicmarko20: I dislike 2 ternaries, and the line is too long.

Kal74:
@OskarStark Sure, but as the behavior is in JS is there a functionnal test suites?

src/Resources/views/Form/datepicker.html.twig
Kal74: All the form type in this bundle have the js embedded in the twig.
If we want to externalize it, it as to be called on page load and by Admin.shared_setup (on modal open) in the AdminBundle.
I could maybe simple share a twig block.

src/Form/Type/BasePickerType.php
```diff
@@ -72,9 +72,12 @@ public function finishView(FormView $view, FormInterface $form, array $options)
         if (isset($options['date_format']) && is_string($options['date_format'])) {
             $format = $options['date_format'];
         } elseif (is_int($format)) {
-            $timeFormat = ($options['dp_pick_time']) ? DateTimeType::DEFAULT_TIME_FORMAT : \IntlDateFormatter::NONE;
+            $timeFormat = \IntlDateFormatter::NONE;
+            if ($options['dp_pick_time']) {
```
kunicmarko20: shouldn't there be a new line before if?

kunicmarko20: strange that flintCI didn't report this?

greg0ire: I don't think there is any CS rule about this…

greg0ire:
Thanks @Kal74 !

kunicmarko20:
Please use squash and merge or tell the developer that provided PR to squash his commits.

greg0ire:
Argh! Sorry!

PR #546, opened 2018-06-12
#545 Fix fputcsv() default escape character

### LINKED ISSUE #545: Warning: fputcsv(): escape must be a character
I made this issue for my pull request, it's already done.

The problem is in an exporter: [default escape value is empty string](https://github.com/sonata-project/SonataCoreBundle/blob/3.x/src/Exporter/Exporter.php#L60) which is illegal for [fputcsv](http://php.net/manual/en/function.fputcsv.php) function:

> **escape_char**
The optional escape_char parameter sets the escape character (one character only).

The error is: `Warning: fputcsv(): escape must be a character`

Version details:
```
composer show --latest 'sonata-project/*' | grep core-bundle
sonata-project/core-bundle               3.9.1  = 3.9.1  Symfony SonataCoreBundle

```

DmitryRomanenko:
I am targeting this branch, because CSV export is broken in it.

Closes #545

### Fixed
- fputcsv() default escape character

## Subject

The problem is in an exporter: [default escape value is empty string](https://github.com/sonata-project/SonataCoreBundle/blob/3.x/src/Exporter/Exporter.php#L60) which is illegal for [fputcsv](http://php.net/manual/en/function.fputcsv.php) function:

> **escape_char**
The optional escape_char parameter sets the escape character (one character only).

The error is: `Warning: fputcsv(): escape must be a character`

Version details:
```
composer show --latest 'sonata-project/*' | grep core-bundle
sonata-project/core-bundle               3.9.1  = 3.9.1  Symfony SonataCoreBundle
```

### DIFF
--- a/src/Exporter/Exporter.php
+++ b/src/Exporter/Exporter.php
@@ -57,7 +57,7 @@ class Exporter
 
                 break;
             case 'csv':
-                $writer = new CsvWriter('php://output', ',', '"', '', true, true);
+                $writer = new CsvWriter('php://output', ',', '"', '\\', true, true);
                 $contentType = 'text/csv';
 
                 break;
--- a/tests/Form/Type/BooleanTypeTest.php
+++ b/tests/Form/Type/BooleanTypeTest.php
@@ -76,7 +76,7 @@ class BooleanTypeTest extends TypeTestCase
 
         $options = $optionResolver->resolve();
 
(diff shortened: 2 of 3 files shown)


greg0ire:
Thanks!

DmitryRomanenko:
Thank you for merging.  
Should we create `3.9.2` with new updates (since February)?

PR #550, opened 2018-06-27
Remove strong dependency on session

tgalopin:
I am targeting 3.x because this PR is BC-compatible and fix a bug with some session handlers.

## Changelog

```markdown
### Added
- Added `FlashMessageRuntime` and `StatusRuntime` to remove strong dependency on the session introduced by Twig extensions
```

## Subject

Two Twig extensions of SonataCoreBundle relies on the session through the flash messages manager. In some contexts such as when sessions are stored in Redis and we try to warmup the cache in CLI without Redis available, this makes the process fails.

This PR fixes this by using Twig runtimes instead of a direct extension to avoid a strong dependency on the flash messages manager.

### DIFF
--- a/src/Resources/config/flash.xml
+++ b/src/Resources/config/flash.xml
@@ -12,9 +12,12 @@
             <argument/>
             <argument/>
         </service>
+        <service id="sonata.core.flashmessage.twig.runtime" class="Sonata\CoreBundle\Twig\Extension\FlashMessageRuntime">
+            <tag name="twig.runtime"/>
+            <argument type="service" id="sonata.core.flashmessage.manager"/>
+        </service>
         <service id="sonata.core.flashmessage.twig.extension" class="%sonata.core.twig.extension.flashmessage.class%" public="true">
             <tag name="twig.extension"/>
-            <argument type="service" id="sonata.core.flashmessage.manager"/>
         </service>
     </services>
 </container>
--- a/src/Twig/Extension/FlashMessageExtension.php
+++ b/src/Twig/Extension/FlashMessageExtension.php
@@ -19,6 +19,7 @@ use Twig\TwigFunction;
  * This is the Sonata core flash message Twig extension.
  *
  * @author Vincent Composieux <composieux@ekino.com>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
  */
 class FlashMessageExtension extends AbstractExtension
 {
@@ -27,16 +28,31 @@ class FlashMessageExtension extends AbstractExtension
      */
     protected $flashManager;
 
-    public function __construct(FlashManager $flashManager)
+    public function __construct(FlashManager $flashManager = null)
     {
         $this->flashManager = $flashManager;
+
+        if ($this->flashManager) {
+            @trigger_error(
+                'Argument "flashManager" in FlashMessageExtension is deprecated since SonataCoreBundle 3.x and will'.
+                ' be removed in 4.0. Use the FlashMessageRuntime instead.',
+                E_USER_DEPRECATED
+            );
+        }
     }
 
     public function getFunctions()
     {
+        if ($this->flashManager) {
+            return [
+                new TwigFunction('sonata_flashmessages_get', [$this, 'getFlashMessages']),
+                new TwigFunction('sonata_flashmessages_types', [$this, 'getFlashMessagesTypes']),
+            ];
+        }
+
         return [
-            new TwigFunction('sonata_flashmessages_get', [$this, 'getFlashMessages']),
-            new TwigFunction('sonata_flashmessages_types', [$this, 'getFlashMessagesTypes']),
+            new TwigFunction('sonata_flashmessages_get', [FlashMessageRuntime::class, 'getFlashMessages']),
+            new TwigFunction('sonata_flashmessages_types', [FlashMessageRuntime::class, 'getFlashMessagesTypes']),
         ];
     }
 
@@ -47,9 +63,17 @@ class FlashMessageExtension extends AbstractExtension
      * @param string $domain Translation domain to use
      *
      * @return string
+     *
+     * @deprecated since 3.x, to be removed in 4.0. Use the FlashMessageRuntime instead.
      */
     public function getFlashMessages($type, $domain = null)
     {
+        @trigger_error(
+            'Method "FlashMessageExtension::getFlashMessages()" is deprecated since SonataCoreBundle 3.x and will'.
+            ' be removed in 4.0. Use the FlashMessageRuntime instead.',
+            E_USER_DEPRECATED
+        );
+
         return $this->flashManager->get($type, $domain);
     }
 
@@ -57,9 +81,17 @@ class FlashMessageExtension extends AbstractExtension
      * Returns flash messages types handled by Sonata core flash manager.
      *
      * @return string
+     *
+     * @deprecated since 3.x, to be removed in 4.0. Use the FlashMessageRuntime instead.
      */
     public function getFlashMessagesTypes()
     {
+        @trigger_error(
+            'Method "FlashMessageExtension::getFlashMessagesTypes()" is deprecated since SonataCoreBundle 3.x and will'.
+            ' be removed in 4.0. Use the FlashMessageRuntime instead.',
+            E_USER_DEPRECATED
+        );
+
         return $this->flashManager->getHandledTypes();
     }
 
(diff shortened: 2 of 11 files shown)


src/Twig/Extension/FlashMessageExtension.php
```diff
@@ -11,58 +11,25 @@
...
+ * @author Titouan Galopin <galopintitouan@gmail.com>
  */
 class FlashMessageExtension extends AbstractExtension
 {
-    /**
-     * @var FlashManager
-     */
-    protected $flashManager;
-
-    public function __construct(FlashManager $flashManager)
```
kunicmarko20: this is a BC Break

src/Twig/Extension/FlashMessageExtension.php
```diff
@@ -11,58 +11,25 @@
...
-    {
-        return $this->flashManager->get($type, $domain);
-    }
-
-    /**
-     * Returns flash messages types handled by Sonata core flash manager.
-     *
-     * @return string
-     */
-    public function getFlashMessagesTypes()
```
kunicmarko20: probably also a BC Break

src/Twig/Extension/StatusExtension.php
```diff
@@ -11,55 +11,21 @@
...
-     */
-    protected $statusServices = [];
-
-    /**
-     * Adds a renderer to the status services list.
-     *
-     * @param StatusClassRendererInterface $renderer
-     */
-    public function addStatusService(StatusClassRendererInterface $renderer)
-    {
```
kunicmarko20: BC break

src/Twig/Extension/StatusExtension.php
```diff
@@ -11,55 +11,21 @@
...
 
-    /**
-     * @param mixed  $object
-     * @param mixed  $statusType
-     * @param string $default
-     *
-     * @return string
-     */
-    public function statusClass($object, $statusType = null, $default = '')
-    {
```
kunicmarko20: same

src/Twig/Extension/FlashMessageExtension.php
kunicmarko20: If I am correct, you can't delete construct and methods, you can change them to call methods from FlashMessageRuntime class, and add deprecations to construct and methods here.

tgalopin:
I fixed BC breaks.

src/Twig/Extension/FlashMessageExtension.php
```diff
@@ -27,16 +28,31 @@ class FlashMessageExtension extends AbstractExtension
...
     protected $flashManager;
 
-    public function __construct(FlashManager $flashManager)
+    public function __construct(FlashManager $flashManager = null)
     {
         $this->flashManager = $flashManager;
+
+        if ($this->flashManager) {
+            @trigger_error(
+                'Argument "flashManager" in FlashMessageExtension is deprecated since SonataCoreBundle 3.11 and will'.
```
kunicmarko20: Just put 3.x, Release Manager updates this.

src/Twig/Extension/FlashMessageExtension.php
```diff
@@ -47,19 +63,35 @@ public function getFunctions()
      * @param string $domain Translation domain to use
      *
      * @return string
+     *
+     * @deprecated since 3.11, to be removed in 4.0. Use the FlashMessageRuntime instead.
```
kunicmarko20: I guess same for the comments, RM updates them. cc @greg0ire

kunicmarko20:
Looks good, can you check travis?

tgalopin:
I fixed the comments and Travis is running.

kunicmarko20:
Still failing, also the upgrade note is needed

src/Twig/Extension/FlashMessageRuntime.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @author Vincent Composieux <composieux@ekino.com>
+ * @author Titouan Galopin <galopintitouan@gmail.com>
+ */
+class FlashMessageRuntime
+{
+    /**
+     * @var FlashManager
+     */
+    protected $flashManager;
```
greg0ire: Please use `final` and `private` as much as you can

src/Twig/Extension/StatusRuntime.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    /**
+     * @var StatusClassRendererInterface[]
+     */
+    protected $statusServices = [];
+
+    /**
+     * Adds a renderer to the status services list.
+     *
+     * @param StatusClassRendererInterface $renderer
```
greg0ire: Please remove line 29 and 30

src/Twig/Extension/StatusRuntime.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * @param mixed  $object
+     * @param mixed  $statusType
+     * @param string $default
+     *
+     * @return string
+     */
+    public function statusClass($object, $statusType = null, $default = '')
+    {
+        /** @var StatusClassRendererInterface $statusService */
```
greg0ire: Please use `assert($statusService instanceof StatusClassRendererInterface) instead

src/Twig/Extension/StatusRuntime.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param StatusClassRendererInterface $renderer
+     */
+    public function addStatusService(StatusClassRendererInterface $renderer)
+    {
+        $this->statusServices[] = $renderer;
+    }
+
+    /**
+     * @param mixed  $object
```
greg0ire: `object $object`, or is it really mixed?

src/Twig/Extension/FlashMessageRuntime.php
tgalopin: I think this would break the migration path as users currently extending StatusExtension/FlashMessageExtension could not extends anything anymore. But I can do it if you want.

src/Twig/Extension/StatusRuntime.php
tgalopin: I don't know, I copy-pasted the code from the extension.

src/Twig/Extension/FlashMessageRuntime.php
greg0ire: I think they can still rely on service decoration, right, so it would make things a bit harder but still possible.

src/Twig/Extension/StatusRuntime.php
greg0ire: Ok,  please keep it that way then :)

src/Resources/config/flash.xml
```diff
@@ -2,6 +2,7 @@
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <parameters>
         <parameter key="sonata.core.flashmessage.manager.class">Sonata\CoreBundle\FlashMessage\FlashManager</parameter>
+        <parameter key="sonata.core.twig.runtime.flashmessage.class">Sonata\CoreBundle\Twig\Extension\FlashMessageRuntime</parameter>
```
core23: Please don't use parameters for services. We should get rid of them.

tgalopin:
I updated the PR and fixed Travis.

tgalopin:
Ping @core23 :)

greg0ire:
Thanks @tgalopin !

OskarStark:
@tgalopin because of #553 can you please add test cases for the new classes StatusRuntime and FlashMessagesRuntime?

Not important, but shall we add the old authors as new authors of these classes, too? IMO not

OskarStark:
@greg0ire do we need to Mark some tests with group legacy here?

greg0ire:
Every test that covers a deprecated path needs to be marked as legacy

tgalopin:
@OskarStark the StatusRuntime is tested inside StatusExtensionTest and I didn't think the FlashMessageRuntime held enough logic (it's just a proxy) to be tested. The extensions themselves are still tested like before.

It seems however that I missed a BC-break, I'd be happy to add a test to fix it once found.

The code of the runtime is a copy-paste from the previous extensions, I don't think it would be wise to remove the old authors, even if the class is new.

PR #559, opened 2018-07-21
Split manager

core23:

## Changelog

```markdown
### Added
- Added `CommandManagerTrait` and `CommandManagerInterface`
- Added `QueryManagerTrait` and `QueryManagerInterface`
```

## Subject

Apply the ISP to create more specific classes.

### DIFF
--- a/src/Model/BaseManager.php
+++ b/src/Model/BaseManager.php
@@ -20,6 +20,8 @@ use Doctrine\Common\Persistence\ObjectRepository;
  */
 abstract class BaseManager implements ManagerInterface
 {
+    use CommandManagerTrait, QueryManagerTrait;
+
     /**
      * @var ManagerRegistry
      */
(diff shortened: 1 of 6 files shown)


src/Model/ManagerInterface.php
```diff
@@ -66,20 +26,11 @@ public function find($id);
...
-    /**
-     * Delete an Entity.
+     * Return the Entity class name.
      *
-     * @param object $entity   The Entity to delete
-     * @param bool   $andFlush Flush the EntityManager after deleting the object?
+     * @return string
      */
-    public function delete($entity, $andFlush = true);
+    public function getClass();
```
kunicmarko20: Probably a BC Break

src/Model/ManagerInterface.php
```diff
@@ -16,48 +16,8 @@
 /**
  * @author Sylvain Deloux <sylvain.deloux@ekino.com>
  */
-interface ManagerInterface
+interface ManagerInterface extends CommandManagerInterface, QueryManagerInterface
```
core23: Maybe we should deprecate this class in general?

All remaining methods expose internal functionality, which is not a good idea.

src/Model/BaseManager.php
```diff
@@ -20,6 +20,8 @@
  */
 abstract class BaseManager implements ManagerInterface
 {
+    use CommandManagerTrait, QueryManagerTrait;
```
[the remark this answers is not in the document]
core23: I was thinking about the stuff symfony was doing with the `Controller` class. We should remove the `BaseManager` with the next major release.

With this step someone could already create own classes with just the needed parts (query or commands) without import the whole other bunch of stuff..

src/Model/ManagerInterface.php
greg0ire: I think so too, that would lead to not needing the traits anymore, people would implement several interfaces in several classes instead of just one. OK for the traits if you feel this is kind of a compulsory step towards that

core23:
Closing this, as we will remove this bundle

PR #563, opened 2018-09-09
Upgrade dev dependencies

greg0ire:


phpunit.xml.dist
```diff
@@ -14,7 +14,6 @@ It's auto-generated by sonata-project/dev-kit package.
          convertWarningsToExceptions="true"
          processIsolation="false"
          stopOnFailure="false"
-         syntaxCheck="false"
```
greg0ire: See https://github.com/sonata-project/dev-kit/pull/416/files

.travis/install_test.sh
```diff
@@ -4,7 +4,7 @@ set -ev
 mkdir --parents "${HOME}/bin"
 
 # PHPUnit install
-wget "https://phar.phpunit.de/phpunit-5.7.phar" --output-document="${HOME}/bin/phpunit"
+wget "https://phar.phpunit.de/phpunit-7.phar" --output-document="${HOME}/bin/phpunit"
```
kunicmarko20: Isn't this now handled by dev-kit?

greg0ire: It is, but that doesn't matter because I am going to handle the conflict in #568

PR #566, opened 2018-09-10
Use Twig namespaces

greg0ire:


src/Test/XliffValidatorTestCase.php
```diff
@@ -11,10 +11,11 @@
 
 namespace Sonata\CoreBundle\Test;
 
+use PHPUnit\Framework\TestCase;
 use Symfony\Component\Translation\Exception\InvalidResourceException;
 use Symfony\Component\Translation\Loader\XliffFileLoader;
 
-abstract class XliffValidatorTestCase extends \PHPUnit\Framework\TestCase
+abstract class XliffValidatorTestCase extends TestCase
```
greg0ire: This is kind of a BC-break for consumers that need to use a version of phpunit that doesn't know about the namespaced version yet…

greg0ire: :thinking: looks like we are fine after all: https://github.com/sebastianbergmann/phpunit/blob/5.7.27/src/ForwardCompatibility/TestCase.php If things break, we will just have to bump the phpunit version.

PR #570, opened 2018-09-13
Setup NPM/Yarn package

soullivaneuh:
:warning: The branch is pushed on the origin repository to make package install tests easier.

## Changelog

```markdown
### Added
- NPM/Yarn package

### Removed
- Imported vendor assets
- Ionicons library (not used)
```

## Subject

This is an experiment for https://github.com/sonata-project/dev-kit/issues/398.

Currently, I only moved the imported vendor to a NPM/Yarn package.

What to do next? Any suggestion is welcomed

- [ ] Move the JS/CSS files to an another folder. Maybe `assets` on the root directory?
- [ ] Adapte the files to be build with webpack/encore with ease and provide the documentation
- [ ] Find a way to provide dist js/css files for people not using webpack? If yes, how?
- [ ] Publish the package. Which version ? 4.0 ? 0.0, waiting a stable release?
- [ ] Anything else?


kunicmarko20:
I think this should be pushed for later as we are currently doing major releases very slowly and this will just make it even slower.

Soullivaneuh:
@kunicmarko20 For that, wait and see. At the reverse, it would worth to release this on 4.0 and not do a major weeks after that...

kunicmarko20:
I agree it would be good, but I don't think we have enough people to do this as we have a lot of bundles.

Soullivaneuh:
@kunicmarko20 According to the dev-kit issue, we only have 7 bundles concerned about the assets problematic.

In any case, that not the question. Let's talk about how we would manage it, we will see for the release later!

kunicmarko20:
> Move the JS/CSS files to an another folder. Maybe assets on the root directory?
Best practice is `Resources/public` why would we change this?

> Find a way to provide dist js/css files for people not using webpack? If yes, how?
Rebuilding of files on every release and add dist to `Resources`

> Publish the package. Which version ? 4.0 ? 0.0, waiting a stable release?
So I did not understand this, I thought we would just move the vendor libs to npm, but you are talking about moving all sonata asset files there? (or I am still not getting it?)

greg0ire:
>  Find a way to provide dist js/css files for people not using webpack? If yes, how?

Maybe misappropriate Composer to achieve this? Or maybe drop that use case? I don't know webpack well, but do you think we could make it so all people would have to do is change their `packages.json`,  run `yarn install; yarn build` and that's all? Wouldn't webpack get installed automatically by Sonata libs? Regarding the releases, it would be great not to have this issue postpone them a lot… maybe set a deadline for this, after which we release, like October the 1st?

Soullivaneuh:
> Best practice is `Resources/public` why would we change this?

It's an open question. Both can be ok I think.

> Rebuilding of files on every release and add dist to `Resources`

I was thinking about post the compiled resources on something like an open-source CDN more than on a the github repo. Or maybe a special repo for this?

> I thought we would just move the vendor libs to npm, but you are talking about moving all sonata asset files there? (or I am still not getting it?)

Why not both? In any case, if we remove the vendor packages for npm, it will force the user to use NPM/Yarn. This is why I'm also talking about built assets, like bootstrap did for example.

> Maybe misappropriate Composer to achieve this? Or maybe drop that use case?

Composer is not the best way, but maybe we can handle both. But for vendor package, NPM/Yarn will be mandatory except if we provided bundled and compiled assets files.

> I don't know webpack well, but do you think we could make it so all people would have to do is change their packages.json, run yarn install; yarn build and that's all? Wouldn't webpack get installed automatically by Sonata libs? 

Yes for the package installation, not for webpack. We must just provide a dependency file like we do for composer and provide our JS/CSS entrypoint.

The compilation must be the choice of the user. He might use gulp/grunt instead of webpack for example. But I would recommend to publish a documentation at least for webpack and, if the community want, accept PR for other tools documentation.

> maybe set a deadline for this, after which we release, like October the 1st?

The deadline before we release regardless of this PR might be a solution, but less than two weeks is quite short IMHO.

kunicmarko20:
So, its been 3 days and no one besides us responded, are we still sure this is possible to be done? Especially in some limited time period?

Soullivaneuh:
@kunicmarko20 I'm not a Sonata employee. :-)

> Please suggest a deadline then.

Before suggesting a deadline, it would be better to have reaction about the merge request itself (see list on body).

kunicmarko20:
I know, and that is my point, no one is and there aren't active people here. I will just shut up now and let you guys decide.

core23:
👍 For removing all asset copies.

As we will deprecate the whole CoreBundle soon, we could also remove the (old) vendor assets and prefer a `package.json` config.

Soullivaneuh:
> we could also remove the (old) vendor assets and prefer a package.json config.

I did a quick tour. Almost all vendor assets are used. Only one was removed.

Thanks for the comment! And what about people not using NPM or Yarn? I don't think we should make this mandatory use it.

kunicmarko20:
I have been thinking, since we are going to split core after the major, does it even make sense to do this now or maybe better to do it for every lib separately?

core23:
IMHO this doensn't make any sense for now, because we will move the `package.json` to the form-extensions repo soon

core23:
We deprecate this project soon, so we should handle the assets stuff in the other projects.

PR #582, opened 2018-11-19
3.12.0

greg0ire:
Just approve


greg0ire:
I was going to say that it is pedantic, but is it really?

PR #586, opened 2018-11-26
Manual merge

greg0ire:


greg0ire:
The build breaks with: 

> TypeError: IntlDateFormatter::__construct() expects parameter 6 to be string, null given

I don't understand why the build would pass on 3.x

greg0ire:
Ah I understand, it's about the "strict types" declaration.

README.md
```diff
@@ -39,3 +39,8 @@ This package is available under the [MIT license](LICENSE).
 [coveralls_stable_link]: https://coveralls.io/github/sonata-project/SonataCoreBundle?branch=3.x
 [coveralls_unstable_badge]: https://coveralls.io/repos/github/sonata-project/SonataCoreBundle/badge.svg?branch=master
 [coveralls_unstable_link]: https://coveralls.io/github/sonata-project/SonataCoreBundle?branch=master
+
+[scrutinizer_stable_badge]: https://scrutinizer-ci.com/g/sonata-project/SonataCoreBundle/badges/quality-score.png?b=3.x
```
OskarStark: this looks wrong

docs/requirements.txt
```diff
@@ -1,3 +1,3 @@
-Sphinx
+Sphinx!=1.8.0
 git+https://github.com/fabpot/sphinx-php.git
-sphinx_rtd_theme
\ No newline at end of file
+sphinx_rtd_theme
```
OskarStark: the dev kit changes should be available as they get executed against `master` separately

greg0ire: What should I do? Won't dev-kit fix anything that goes wrong?

README.md
greg0ire: Should it be master?

docs/requirements.txt
OskarStark: Ofc it will, but only when it gets executed again

README.md
OskarStark: yes and it is currently available, maybe your branch was outdated? 🤔

OskarStark: ah sorry, no it isn't available in `master`, but yes it should be `master` in the URL

greg0ire: But it says "stable", and master is unstable.

OskarStark: argh you are right, sorry for confusing.

Everything is fine 👍

PR #609, opened 2018-12-12
Fix forms namespace

danut007ro:
## Fix forms namespace in form_types.xml

After moving the namespace from `Sonata\CoreBundle\Form` to `Sonata\Form` the form_types.xml and date.xml still pointed to old namespace.

I am targeting this branch, because it is a bug fix.

## Changelog

Fix namespace in form_types.xml and date.xml

### DIFF
--- a/src/CoreBundle/Resources/config/form_types.xml
+++ b/src/CoreBundle/Resources/config/form_types.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <service id="sonata.core.form.type.array" class="Sonata\CoreBundle\Form\Type\ImmutableArrayType" public="true">
+        <service id="sonata.core.form.type.array" class="Sonata\Form\Type\ImmutableArrayType" public="true">
             <tag name="form.type" alias="sonata_type_immutable_array"/>
         </service>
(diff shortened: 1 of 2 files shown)


greg0ire:
Can you please test #610? I think it might solve your issue.

danut007ro:
> Can you please test #610? I think it might solve your issue.

Sorry it didn't fixed it, and I found another issue where `use Sonata\CoreBundle\Model\Adapter\AdapterInterface;` in TwigExtension.php should be replaced with `use Sonata\Doctrine\Adapter\AdapterInterface;`.

I'm on php7.3 but this shouldn't matter I think.

greg0ire:
There are some classes I could not touch, the base picker type related ones, are those the one you are still having issues with?

danut007ro:
@greg0ire yes.
But I modified all form types in `form_types.xml`.

I tried and I can update this pull request to modify only `DateTimePickerType`, `DateRangePickerType` and `DateTimeRangePickerType` and it works without needed modification for `sonata.core.date.moment_format_converter`.

Let me know if I should update to this.

src/CoreBundle/Resources/config/form_types.xml
```diff
@@ -23,12 +23,12 @@
             <tag name="form.type" alias="sonata_type_datetime_range"/>
             <argument type="service" id="translator"/>
         </service>
-        <service id="sonata.core.form.type.date_picker" class="Sonata\CoreBundle\Form\Type\DatePickerType" public="true">
+        <service id="sonata.core.form.type.date_picker" class="Sonata\Form\Type\DatePickerType" public="true">
```
core23: I'm not sure if we should do this. If someone want's to use the old form type, there could be problems

core23:
Not after merging https://github.com/sonata-project/SonataCoreBundle/pull/629

PR #610, opened 2018-12-12
Apply aliasing over extension wherever we can

greg0ire:
Refs #609
## Subject

Several issues have been reported because of type hinting issues: code
would type hint against the extending class, and receive a class of the
extended type, for instance. Aliasing solves that kind of issue by
making both types equivalent.

## Changelog

```markdown
### Fixed
- several crashes about type hint issues
```

Note that I couldn't do it for some classes extending `BasePickerType`, not sure why but there was an error about too few arguments…


greg0ire:
Apparently, it does not solve every issue, see #609…

PR #623, opened 2018-12-16
Trigger autoload for legacy types

kunicmarko20:
## Subject

By loading the deprecated types, we make sure the class alias they
contain exist, so that code type hinting against the old type accepts
the new. This is necessary because type hinting does not trigger
autoload, since it usually does not need to.

I am targeting this branch, because this is a fix.

## Changelog

```markdown
### Fixed
- crashes when a class type hints against an old type, but receives a new one
```

Resolves:

```
Argument 1 passed to Sonata\AdminBundle\Admin\AbstractAdmin::Sonata\AdminBundle\Admin\{closure}() must be an instance of Sonata\CoreBundle\Validator\ErrorElement, instance of Sonata\Form\Validator\ErrorElement given, called in /Storage/www/sonata/sonata-test/lib/vendor/sonata-project/core-bundle/src/Form/Validator/InlineValidator.php on line 53
```

### DIFF
--- a/src/CoreBundle/Model/Adapter/AdapterChain.php
+++ b/src/CoreBundle/Model/Adapter/AdapterChain.php
@@ -11,11 +11,13 @@
 
 namespace Sonata\CoreBundle\Model\Adapter;
 
-@trigger_error(
-    'The '.__NAMESPACE__.'\AdapterInterface class is deprecated since version 3.x and will be removed in 4.0.'
-    .' Use Sonata\Doctrine\Adapter\AdapterChain instead.',
-    E_USER_DEPRECATED
-);
+if (!class_exists(\Sonata\Doctrine\Adapter\AdapterChain::class, false)) {
+    @trigger_error(
+        'The '.__NAMESPACE__.'\AdapterChain class is deprecated since version 3.x and will be removed in 4.0.'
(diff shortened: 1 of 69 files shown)


greg0ire:
Argh, my solution is wrong, because then the deprecations are triggered :man_facepalming: This is why the build is failing. I know how to to fix it though: call `class_exists(…, false);`, then do the alias.

greg0ire:
Ok, the build should pass now.

src/CoreBundle/Model/Adapter/AdapterChain.php
```diff
@@ -11,11 +11,13 @@
 
 namespace Sonata\CoreBundle\Model\Adapter;
 
-@trigger_error(
-    'The '.__NAMESPACE__.'\AdapterInterface class is deprecated since version 3.x and will be removed in 4.0.'
```
greg0ire: Yes, there were copy/paste issues with `AdapterInterface`.

greg0ire:
The only issue with this approach is that people will get fewer deprecations, for instance if they use the new type once, then use the old types, they will not get deprecations anymore I'm afraid. Can be improved upon later if it really turns out to be an issue, right now the goal is to fix the crashes.

kunicmarko20:
Deprecations are better than breaks

PR #633, opened 2018-12-19
Use the correct function to trigger autoload

greg0ire:
I think this must be pedantic because there is no way you could know it
is actually an interface before loading the file.


kunicmarko20:
shouldn't this be a patch?

OskarStark:
Thanks @greg0ire 👍 

Good job!

PR #638, opened 2018-12-22
Add isImageAvailable method to Metadata

jo66:
## Subject

This PR will allow use of default mosaic background (see [admin bundle PR #5366](https://github.com/sonata-project/SonataAdminBundle/pull/5366))

I am targeting this branch, because it's feature.

## Changelog

```markdown
### Added
- Added `isImageAvailable` method to `Metadata` class to allow use of default mosaic background
```

## To do

- [x] Update the tests


src/CoreBundle/Model/MetadataInterface.php
```diff
@@ -33,6 +33,11 @@ public function getDescription();
      */
     public function getImage();
 
+    /**
+     * @return boolean
+     */
+    public function isImageAvailable();
```
kunicmarko20: bc break

jo66: Removed change on interface

src/CoreBundle/Model/Metadata.php
```diff
@@ -60,6 +65,7 @@ public function __construct($title, $description = null, $image = null, $domain
...
+        $this->isImageAvailable = null !== $image;
```
greg0ire: No need to store this in a field IMO: computing it every time should be performant enough

jo66: Yes, but as image will be initialized with default value, how to know if it's default or not after initialization? To check image value against default one will be less performant I guess...

src/CoreBundle/Model/Metadata.php
```diff
@@ -39,6 +39,11 @@ class Metadata implements MetadataInterface
      */
     protected $image;
 
+    /**
+     * @var boolean
+     */
+    private $isImageAvailable;
```
core23: This class is deprecated, we shouldn't add any features to this class anymore

jo66: In this case admins that still use it won't benefit from mosaic background... Not a problem? Should I add a deprecation notice about it or just close PR?

src/CoreBundle/Model/Metadata.php
greg0ire: You can store the default value in a constant and do the comparison against it, performance will really not be an issue here IMO.

jo66: Done

tests/CoreBundle/Model/MetadataTest.php
```diff
@@ -48,12 +48,22 @@ public function testGetters()
...
     {
         $metadata = new Metadata('title', 'description', false);
         $this->assertFalse($metadata->getImage());
     }
+
+    public function testIsImageAvailable()
+    {
+        $metadata = new Metadata('title', 'description');
+        $this->assertFalse($metadata->isImageAvailable());
+        $metadata = new Metadata('title', 'description', false);
```
OskarStark: Pleas use a dataProvider instead

jo66: Done.

core23:
We've added a new `Metadata` object to the AdminBundle. So we could close this one, as it is deprecated and have no special domain context.

jo66:
@core23 Thanks for this, I can now use it for default background in admin bundle :)

PR #647, opened 2019-01-12
Invert condition

greg0ire:
## Subject

If the type exists, then we might be triggerring the autoload from it,
and that should not trigger a deprecation.

## Changelog

```markdown
### Fixed
- deprecation wrongly triggered when using the new interface
```


greg0ire:
I want to do a release, and this is a no-brainer, so let's merge.

PR #660, opened 2019-01-26
fix risky tests

OskarStark:
an alternative could be to use @doesNotPerformAssertions annotation
on this testcase but this one is more straight forward.

After merging this, only the test on `--prefer-lowest`-builds, depending on `XliffValidatorTestCase` will be marked as "risky".


greg0ire:
> After merging this, only the test on --prefer-lowest-builds, depending on XliffValidatorTestCase will be marked as "risky".

Unless we require a recent version of the core bundle, but it's not worth the effort IMO :P

OskarStark:
> Unless we require a recent version of the core bundle, but it's not worth the effort IMO 

exactly

PR #679, opened 2019-03-04
Inject adapters on the new doctrine service name

jordisala1991:
## Subject

I am targeting this branch, because this should be BC.

## Changelog

```markdown
### Fixed
- Adapters are not being injected on the adapter chain.
- `sonata_urlsafeid` twig filter is working again
```

### DIFF
--- a/src/CoreBundle/DependencyInjection/Compiler/AdapterCompilerPass.php
+++ b/src/CoreBundle/DependencyInjection/Compiler/AdapterCompilerPass.php
@@ -24,22 +24,22 @@ class AdapterCompilerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        if (!$container->has('sonata.core.model.adapter.chain')) {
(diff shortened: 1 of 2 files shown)


jordisala1991:
Does it makes sense to make this change here or should we move this compiler pass to the sonata-doctrine-extensions, where the chain lives:

https://github.com/sonata-project/sonata-doctrine-extensions/blob/5453bf15b2811d0c3249442897bcc50c67c91d57/src/Bridge/Symfony/Resources/config/doctrine_orm.xml

core23:
👍 for the doctrine repo

The core repo should die asap

jordisala1991:
ohhh... I see... if the doctrine-extensions repo is not enabled, then this works:

https://github.com/sonata-project/SonataCoreBundle/blob/3.x/src/CoreBundle/DependencyInjection/SonataCoreExtension.php#L80

So the compiler pass should be created on the sonata-doctrine-extensions

PR #685, opened 2019-03-15
Removed FlashBag translation

### LINKED ISSUE #345: Duplicate translation in flash messages
| Question | Answer |
| --- | --- |
| Bundle version | 3.1.1 |
| Symfony version | 2.8.4 |
| php version | 5.5 |
## Subject

The flash messages are translated multiple times.
## Steps to reproduce

Add a new (translated) message to the `FlashBag`.
## Expected results

No translation is done.
## Actual results

bildschirmfoto 2016-09-18 um 08 28 14

## Solution

We should remove the translation in the `FlashManager, because that's no good way to handle missing translation of the developer.

core23:
## Subject

The translation should be done  before calling the symfony`FlashBag::add` method. This is the only way to translate messages with parameters.

Closes #345 

## Changelog

```markdown
### Removed
- Removed FlashBag translation
```


kunicmarko20:
Thanks @core23 !

PR #719, opened 2019-11-27
3.17.1

greg0ire:


phansys (dismissed):

greg0ire:
The build fails! Help @core23 :pray:

greg0ire:
That has to do with sf 4.4, blacklisting it for all the component fixes the build, I'm going to try to see if blacklisting one of the components is enough

composer.json
```diff
@@ -23,7 +23,7 @@
         "sonata-project/datagrid-bundle": "^2.0",
         "sonata-project/doctrine-extensions": "^1.1",
         "symfony/config": "^2.8 || ^3.2 || ^4.0",
-        "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
+        "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0,!=4.4.0",
```
phansys, line 26: Shouldn't be better to add `4.4.0` as conflicting version?

core23 (approved):

composer.json
greg0ire: IMO It's the same but you have all the constraints in the same place.

phansys: I agree with that point. Just that using the "conflict" section seems more explicit or semantically correct to me (if in any case we would remove this dependency, but that will not happen because this package is abandoned).
:+1:

phansys (approved):

PR #722, opened 2019-11-28
Allow to use symfony/twig-bridge 4.4.0

fbourigault:
This allow to use `symfony/twig-bridge` `4.4.0`.

Tests failed because tests depends on code that is now no longer shipped with dist because of the new export-ignore rules.

I need to solve the translator mess to get something working.

PR #723, opened 2019-11-28
Steal Test fixture class from Twig bridge

greg0ire:
## Subject

The bridge archives have recently been modified so that they no longer
contain tests. We were relying on one of the fixtures that came with
thes tests. Copying the fixture to our project fixes the issue.
Allows using the twig bridge v4.4

See https://github.com/symfony/twig-bridge/commit/3b24df50a23d2a11d459d9e31dc4cffb8f7caf

Closes #722 

## Changelog

```markdown
### Added
- Compatibility with `symfony/twig-bridge` 4.4
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "symfony/property-access": "^2.8 || ^3.2 || ^4.0",
         "symfony/security-csrf": "^2.8 || ^3.2 || ^4.0",
         "symfony/translation": "^2.8 || ^3.2 || ^4.0",
-        "symfony/twig-bridge": "^2.8 || ^3.2 || ^4.0,!= 4.4.0",
+        "symfony/twig-bridge": "^2.8 || ^3.2 || ^4.0",
         "symfony/validator": "^2.8 || ^3.2 || ^4.0",
         "twig/extensions": "^1.5",
         "twig/twig": "^1.34 || ^2.0"
--- a/src/Form/Fixtures/StubFilesystemLoader.php
+++ b/src/Form/Fixtures/StubFilesystemLoader.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\Form\Fixtures;
+
+use Twig\Loader\FilesystemLoader;
+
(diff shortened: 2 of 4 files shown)


core23 (dismissed):

greg0ire:
1. I forgot one file
2. Now I understand what mess @fbourigault was talking about… there is an interface that's present in `symfony/contracts`, which is a recent package…

greg0ire:
Stand back, I'm trying some black magic

OskarStark:
Looks like it’s working 😈

core23 (dismissed):

fbourigault:
IMHO, we have to conflict with symfony/translation-contracts >= 2.0 because of the added type hints.

greg0ire:
Also, the class that relies on these fixtures is not in `tests`, it's in `src`… I have to move the fixtures in `src` otherwise it will crash since we blacklist `tests` from our archives too.

greg0ire:
@fbourigault actually, there is no conflict as long as people use php 7.2: https://3v4l.org/Kqeo7

greg0ire:
I pushed with the conflict, then we can do a release, and if we want we can do another release that requires 7.2 and removes the conflict after that.

core23 (dismissed):

fbourigault (dismissed):

composer.json
```diff
@@ -33,13 +33,14 @@
...
-        "symfony/twig-bridge": "^2.8 || ^3.2 || ^4.0,!= 4.4.0",
+        "symfony/twig-bridge": "^2.8 || ^3.2 || ^4.0",
         "symfony/validator": "^2.8 || ^3.2 || ^4.0",
         "twig/extensions": "^1.5",
         "twig/twig": "^1.34 || ^2.0"
     },
     "conflict": {
-        "sonata-project/exporter": "< 1.11.0"
+        "sonata-project/exporter": "< 1.11.0",
+        "symfony/translation-contracts": ">=2.0.0"
```
fbourigault: Since `symfony/translation-contracts` requires PHP 7.2, this should not be necessary ?

fbourigault (commented):
Since symfony-contracts v2.0.0 requires PHP 7.2,

composer.json
greg0ire: Oh you're correct, that's great, let me change that :)

core23 (approved):

phansys (approved):

core23:
Changed the label to `patch` as we don't release any new minor version

greg0ire:
I think we are supposed to use `patch` anyway, but it doesn't make sense to me, that's why I did not say anything regarding the label for the seo bundle.

PR #732, opened 2020-03-23
Symfony 5 support

jorrit:
## Subject

Although the core bundle is deprecated, it may be hard to remove it as a dependency without breaking BC. This means that in order to enable upgrading bundles depending on the core bundle to update to Symfony 5, Core bundle must be upgraded to Symfony 5 too. 

Here I'll try to see if this is possible.

I am targeting this branch, because I want to offer Symfony 5 support for depending modules without a major upgrade.

## Changelog

```markdown
### Added
- Added support for Symfony 5.
```

## To do

- [x] Fix format/html5 test error
- [x] Fix visitInteger mock error
- [x] Fix `--prefer-lowest` error in `BaseDoctrineORMSerializationType`


core23 (changes requested):
Blocked for now. We have to discuss if we want this for this deprecated project

VincentLanglet:
I think that you'll need to fix this issue https://github.com/sonata-project/SonataCoreBundle/issues/731

jorrit:
@core23 I completely understand that merging this PR is subject to further discussion. I just wanted to see if it is possible at all.

jorrit:
@core23 Do you think it would be possible to get the latest Dev-Kit files in Sonata Core 3.x?

wbloszyk:
@jorrit I working on #741 and https://github.com/sonata-project/dev-kit/issues/697.

This mean CoreBundle will be drop in some steps:
- upgrade CoreBundle to 3.20,
- remove usage in own code,
- add twig extensions bridge,
- drop CoreBundle,
- upgrade extensions to 1.x (0.x should be use only for drop CoreBundle)

This will allow add Symfony 5 support in Sonata 3. I think you can close this PR.

PR #382, opened 2017-02-02
added dependency for Sonata\CoreBundle\ModelPageableManagerInterface

OskarStark:
I am targetting this branch, because this is BC.

## Changelog

```markdown
### Added
- `sonata-project/datagrid-bundle` dependency for `Sonata\CoreBundle\ModelPageableManagerInterface`
```

## Subject

IMO the direct dependency was missing

screenshot 2017-02-02 09 30 01

PR #735, opened 2020-04-06
Remove DatagridBundle dependency

VincentLanglet:
## Subject

I am targeting this branch, because BC.

The coreBundle does not use a lot the datagrid bundle but it's require in the `composer.json`.

I would like to require the datagridBundle 3.x in the adminBundle, in order to deprecate the Datagrid directory in favor of classes in the datagridBundle. But the adminBundle actually require the coreBundle which require the datagridBundle 2.x.

## Changelog

```markdown
### Remove
- Remove SonataDatagridBundle dependency
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,6 @@
     "require": {
         "php": "^7.2",
         "cocur/slugify": "^1.4 || ^2.0 || ^3.0 || ^4.0",
-        "sonata-project/datagrid-bundle": "^2.0 || ^3.0",
         "sonata-project/doctrine-extensions": "^1.1",
         "symfony/config": "^4.3",
         "symfony/dependency-injection": "^4.3",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "require": {
         "php": "^7.2",
         "cocur/slugify": "^1.4 || ^2.0 || ^3.0 || ^4.0",
-        "sonata-project/datagrid-bundle": "^2.0",
+        "sonata-project/datagrid-bundle": "^2.0 || ^3.0",
         "sonata-project/doctrine-extensions": "^1.1",
         "symfony/config": "^4.3",
         "symfony/dependency-injection": "^4.3",


greg0ire (dismissed):

core23:
I thought this bundle is deprecated. Can't we replace the datagrid with the 5.x release of the admin bundle? We should finish the following major release instead of deprecating too much.

VincentLanglet:
> I thought this bundle is deprecated. Can't we replace the datagrid with the 5.x release of the admin bundle? We should finish the following major release instead of deprecating too much.

The interface in the SonataAdmin and the SonataDatagrid bundles start to be different because every PR are only made on one of them. For example you added generics to the Pager in the datagridBundle but didn't in the adminBundle. The more we wait, the harder it will be.

For example, if this PR https://github.com/sonata-project/SonataAdminBundle/pull/6029 would have been merged, the Datagrid would have need the `FieldDescriptionInterface` which is actually in the adminBundle. So it would be impossible to require the datagridBundle in the adminBundle because of circular dependency.

I would first to make common classes in datagridBundle and adminBundle consistent before it'll be too hard.

core23:
> The more we wait, the harder it will be.

That's exactly the reason why we still have the admin 3.x release, because we wait too long and want to include too many cool features / cleanup.

Back to topic: I'm not sure if we ever use the datagrid bundle inside this repo. Can't we just remove the dependency?

core23 (approved):

greg0ire:
> I'm not sure if we ever use the datagrid bundle inside this repo.

Apparently we did: https://github.com/sonata-project/SonataCoreBundle/pull/382

VincentLanglet:
> > I'm not sure if we ever use the datagrid bundle inside this repo.
> 
> Apparently we did: #382

Since the implementation was removed and deprecated, the dependency can be remove now then. Is it ok for you @greg0ire ? :)

core23:
This commit removes the datagrid: https://github.com/sonata-project/SonataCoreBundle/commit/c2941d1201fb43777d97709712be9e73f323f424

phansys (approved):

greg0ire (approved):
> Is it ok for you @greg0ire ? :)

Now that we understand everything, yes :)

greg0ire:
Thanks @VincentLanglet !

VincentLanglet:
Thanks ! Can you make a release @greg0ire ? :)

Issue #741, opened 2020-05-21
Migrate CoreBundle to extensions


wbloszyk:
## Feature Request

We must consider dropping the CoreBundle in two ways.

## In sonata-project bundles
Some CoreBundle classes extends *-extensions classes. We can not change CoreBundle to *-extensions in sonata-project bundles. We should change `extends` to `class_alias`. This will allow change API and drop CoreBundle in sonata-project bundles.

Issue for that is here https://github.com/sonata-project/dev-kit/issues/697

## In users projects
User must get information about deprecated in IDE. We must deprecated all CoreBundle classes. Some feature like FormHelper must be drop too becouse form type mapping is not support from 3.0 ?? We should also add "Bridge" to Form and Twig. This will allow resolve release problem. We will release CoreBundle, FormBundle and TwigBundle in one package:
```json
 "sonata-project/core-bundle ":  "3.19"
```
and this package will be equal to next release:
```json
 "sonata-project/core-bundle ":  "3.20",
 "sonata-project/form-extensions": "0.1",
 "sonata-project/twig-extensions": "0.1"
```

I prepare v3.19 in 5 RPs:
- [x] [Add deprecations](https://github.com/sonata-project/SonataCoreBundle/pull/742)
- [x] [Add SonataFormBundle (bridge)](https://github.com/sonata-project/SonataCoreBundle/pull/743)
- [x] [Add SonataTwigBunde (bridge)](https://github.com/sonata-project/SonataCoreBundle/pull/744)
- [x] [Add class_aliases](https://github.com/sonata-project/SonataCoreBundle/pull/745)
- [x] [Fix pack with tests ](https://github.com/sonata-project/SonataCoreBundle/pull/746)

CoreBundle v3.20:
- [x] https://github.com/sonata-project/form-extensions/pull/92
- [x] https://github.com/sonata-project/twig-extensions/pull/80
- [x] https://github.com/sonata-project/SonataCoreBundle/pull/752

wbloszyk:
I finish this issue. RTM.

jordisala1991:
This one can be closed right? we can continue on the dev kit issue

PR #742, opened 2020-05-21
Add deprecations

wbloszyk:
## Subject

I am targeting this branch, because it is BC.

Part of #741 

## Changelog

```markdown
## Deprecated
- all CoreBundle class and services
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "twig/twig": "^1.34 || ^2.0"
     },
     "conflict": {
-        "sonata-project/exporter": "< 1.11.0"
+        "sonata-project/exporter": "< 1.11"
     },
     "require-dev": {
         "doctrine/orm": "^2.4",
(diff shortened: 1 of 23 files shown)


greg0ire:
Maybe we should merge this after all code in our packages is migrated?

wbloszyk:
@greg0ire Please read #741  one more time. This way will be much better.
I have all this PRs prepared. Just check it and after all release. We can drop CoreBundle in Sonata 3. After this change we make https://github.com/sonata-project/dev-kit/issues/697. Next will be https://github.com/sonata-project/SonataBlockBundle/issues/700. Then we can focus how use 0.x extensions in Sonata 4.

greg0ire:
> Maybe we should merge this after all code in our packages is migrated?

What I mean is maybe we should add deprecations only after sonata-project/dev-kit#697 is done. That way end users do not get deprecations they cannot fix.

wbloszyk:
> > Maybe we should merge this after all code in our packages is migrated?
> 
> What I mean is maybe we should add deprecations only after [sonata-project/dev-kit#697](https://github.com/sonata-project/dev-kit/issues/697) is done. That way end users do not get deprecations they cannot fix.

I understand what you mean. Anyway we cant done [sonata-project/dev-kit#697](https://github.com/sonata-project/dev-kit/issues/697) before #741. We need stable 0.x extensions and this is the easiest way to do it. We can add this deprecations after both issues but for now CoreBundle also have some deprecations which user couldn't fix. I think we should not avoid add new deprecations if in this way we can resolve this BIG CORE problem.

greg0ire:
It's it's much harder to add the deprecations later, don't do it. It's better DX but I'd rather have this done than discourage you :sweat_smile:

greg0ire (approved):

core23 (approved):

PR #745, opened 2020-05-22
Add class aliases

wbloszyk:
## Subject

I am targeting this branch, because it is BC.

Part of #741


core23 (approved):
Looks okay for me.

Do you have to add the new bridges to your `bundles.php`(e.g. SonataTwigBundle)? If yes, we should not (yet) merge this. There should be no manual work when updating this library via composer.

wbloszyk:
Yes, bridges are required. I add something like this:
```php
if (!isset($bundles['SonataFormBundle'])) {
            throw new NotRegisterBundleException('Please register SonataFormBundle. Add the following line in the config/bundles.php:
            <?
            return [
                    ...
                  Sonata\Form\Bridge\Symfony\SonataFormBundle::class => [\'all\' => true];
             ];
            ');
        }

        if (!isset($bundles['SonataTwigBundle'])) {
            throw new NotRegisterBundleException('Please register SonataTwigBundle. Add the following line in the config/bundles.php:
            <?
            return [
                    ...
                  Sonata\Twig\Bridge\Symfony\SonataTwigBundle::class => [\'all\' => true];
             ];
            ');
        }
```
We can automate that by recipes.

wbloszyk:
Ok, This is stupid idea. Exception in BundleExtension will crash Travis. I wll load additional (bridge) config instead. This 4 PRs are RTM. I will try finish last today. Push my sandbox to reproduce Sonata 3 without core and open PR for  https://github.com/sonata-project/dev-kit/issues/697.

wbloszyk:
> Looks okay for me.
> 
> Do you have to add the new bridges to your `bundles.php`(e.g. SonataTwigBundle)? If yes, we should not (yet) merge this. There should be no manual work when updating this library via composer.

#741 Is ready. It's allow use `sonata.form.*` and `sonata.twig.*' without register bridge. Then CoreBundle config is inject to services. If you load then form- and twig- extensions config is injected.

greg0ire:
Thanks @wbloszyk !

PR #751, opened 2020-05-26
Improve core before split - CoreBundle 3.19.2

wbloszyk:
## Subject

Allow register bridges by move aliases to compiler pass. Override CoreBundle config will throw exception about move this config to registred bridges.

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Changed
- change register aliases from bundle configuration to compiler pass
### Fixed
- fixed register CoreBundle with form and twig bridges 
```

### This PR should be release as 3.19.2


VincentLanglet (commented):
Why should we release this as 3.19.2 and not 3.20 ?

This seems more like a release than a patch to me

wbloszyk:
It can be release as 3.20 too. I will change reqiurements in feature PRs then.

src/CoreBundle/DependencyInjection/Compiler/FormCompilerPass.php
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Sonata\CoreBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @deprecated since sonata-project/core-bundle 3.19, to be removed in 4.0.
```
VincentLanglet:
```suggestion
 * @deprecated since sonata-project/core-bundle 3.20, to be removed in 4.0.
```
Or
```suggestion
 * @deprecated since sonata-project/core-bundle 3.x, to be removed in 4.0.
```
And we'll replace `3.x` by the right release number.

Same for others.

wbloszyk:
@VincentLanglet I check something. This PR replace creating alias from xml files to compiler pass. It can be trait as patch or minor. Some change is already merge.  Better keep it as 3.19.2 to avoid fix it in production.

src/CoreBundle/DependencyInjection/Compiler/FormCompilerPass.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public function registerDateAliases(ContainerBuilder $container)
+    {
+        $momentFormatConventerAlias = $container
+            ->setAlias('sonata.core.date.moment_format_converter', 'sonata.form.date.moment_format_converter')
+            ->setPublic(true)
+            ->setDeprecated(
+                true,
+                'The "%service_id%" service is deprecated since sonata-project/core-bundle 3.19 and will be removed in 4.0. Use "%alias_id%" instead.'
```
jordisala1991: This is what this message is producing: `The "%service_id%" service is deprecated since sonata-project/core-bundle 3.19 and will be removed in 4.0. Use "sonata.core.date.moment_format_converter" instead. It is being referenced by the "sonata.core.form.type.date_picker_legacy" service.`

I think it should be: `The "%alias_id%" service is deprecated since sonata-project/core-bundle 3.19 and will be removed in 4.0. Use "sonata.form.date.moment_format_converter" instead.`
And this will produce: `The "sonata.core.date.moment_format_converter" service is deprecated since sonata-project/core-bundle 3.19 and will be removed in 4.0. Use "sonata.form.date.moment_format_converter" instead. It is being referenced by the "sonata.core.form.type.date_picker_legacy" service.`

Reference: https://travis-ci.org/github/sonata-project/SonataAdminBundle/jobs/693376005#L791

It applies for all messages

wbloszyk: This message are correct in xml and should be good in PHP too. I will check it when i find free time.

jordisala1991:
I did a couple of PR on your fork to improve this PR, feel free to merge them or use portions of code to help finish this PR.

wbloszyk:
I have long weekend. I will back to home tomorrow. Some change look nice.

VincentLanglet:
> @VincentLanglet I check something. This PR replace creating alias from xml files to compiler pass. It can be trait as patch or minor. Some change is already merge. Better keep it as 3.19.2 to avoid fix it in production.

I don't understand.

Patch is for bugfix.
Minor is for new feature.

Are you fixing a bug ?
I don't think so. You're adding new CompilerPass and you're adding deprecation.

If someone require 3.19.x. Updating automatically from 3.19.1 to 3.19.2 could break a CI because of the deprecation for example. There is certainly other possible impact.

Even if we're deprecating the bundle, we should still respect the SemVer.

wbloszyk:
I adding fix, becouse bridges exist and can't be register. It is not new feature too

src/CoreBundle/DependencyInjection/Compiler/FormCompilerPass.php
```diff
@@ -0,0 +1,158 @@
...
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @deprecated since sonata-project/core-bundle 3.x, to be removed in 4.0.
+ */
+class FormCompilerPass implements CompilerPassInterface
```
ChristianGripp: Please make this final and reduce visibility to private, if possible.

core23 (changes requested):

jordisala1991 (approved):

VincentLanglet (approved):

src/CoreBundle/DependencyInjection/Compiler/FormCompilerPass.php
```diff
@@ -0,0 +1,158 @@
...
+        $defaultSerializer = [
+                'formats' => [
+                    0 => 'json',
+                    1 => 'xml',
+                    2 => 'yml',
+                ],
+            ];
+
+        if ($container->getParameter('sonata.core.form') !== $defaultForm) {
+            throw new \Exception('Move bundle config from sonata_core.form to sonata_form.form');
```
ChristianGripp, line 151: Sure we should throw an exception? A deprecation message should be preferred for a smooth upgrade path.

core23 (changes requested):

src/CoreBundle/DependencyInjection/Compiler/FormCompilerPass.php
wbloszyk: CoreBundle register own form and twig service based on core config when user do not register bridges.

When user register bridges then form and twig service will be loaded from bridges and bridges config, so config must be replace too. 

Register new bundles/bridges if equal to major change. Throw exception is awesome solution here becouse we can force user to do something. Mybe we should improve this  message?

For example:
"Your register form-extensions. From now registred SonataCoreBundle will be using form-extensions configuration. Please move sonata_core.form to sonata_form.form and keep sonata_core.form clear(empty)."

ChristianGripp: > Throw exception is awesome solution here becouse we can force user to do something.

We should not *force* the user to do something when updating to a minor version

wbloszyk: And we don't. We force the user to do something in two cases:
1. Update CoreBundle and register new bridges from CoreBundle 
2. Update CoreBundle and register bridges from extensions 0.x

If user update bundles without register this bridges then we do not throw this exception. So we force the user to do something when new bridges will be register by the user. It is not update minor but part of bridge configuration.

core23 (approved):

src/CoreBundle/DependencyInjection/Compiler/FormCompilerPass.php
ChristianGripp: I missed this line:
https://github.com/sonata-project/SonataCoreBundle/pull/751/files#diff-2f96961c5d13c1ba7ad8ea345be1a57eR131

jordisala1991:
Thank you @wbloszyk !

PR #759, opened 2021-09-29
Update homepage

VincentLanglet:


composer.json
```diff
@@ -5,7 +5,7 @@
     "keywords": [
         "sonata"
     ],
-    "homepage": "https://sonata-project.org/bundles/core",
+    "homepage": "https://github.com/sonata-project/SonataCoreBundle",
```
wbloszyk, line 8: We should remove documentation from `README.md` or add it to docs.sonata-project.org in `Abandoned/Deprecated` section.

wbloszyk (changes requested):

composer.json
VincentLanglet: This bundle was deprecated/abandoned one year ago and is not even updated with the dev kit.
Anyway, this should just be archived instead.

